PR #2, opened 2021-03-21
Clean-up AbstractClassResolver

JesusValera:
## 📚 Description

Clean the `AbstractClassResolver`.
I also check the other project class and seems fine. 😄 

## 🔖 Changes

- Rearrange class
- Remove useless methods (`findCacheKey()`, `resolveClassName ()`)
- Return `void` on `setCallerObject()`
- `getCacheKey()` L26 returns `string`, remove the `$cacheKey !== null` condition on L28 & L32


Chemaclass (approved):
Thanks!

PR #5, opened 2021-03-26
Implementing a Locator to get the Facades from the DependencyProvider

Chemaclass:
## 📚 Description

The Locator will avoid creating a new instance of each Facade. They will be reused as singletons via all different modules which use the Container->getLocator() in their DependecyProvider.

## 🔖 Changes

- Add a Locator inside the Container.


JesusValera (approved):
😍

PR #14, opened 2021-04-04
Create config directory

Chemaclass:
## 📚 Description

Config will load all files under the "config/" directory and the "/config/local.php" the last one


JesusValera (approved):
I find this feature super interesting! 💥

PR #17, opened 2021-04-17
Decouple the "maker templates" from the domain

Chemaclass:
## 🔖 Changes

- Decouple the maker templates from the domain. Inject them instead.

I will create a follow-up PR refactoring the `CommandArgumentsParser` in order to achieve what we talked a few days ago about the "1 argument only needed":
> php gacela make:module Gacela/TestModule

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 /.idea/
 /.vscode/
 /vendor/
+/TestModule/
 .phpunit.*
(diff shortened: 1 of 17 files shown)


JesusValera (approved):
Much more decoupled now 👏🏼

Tito-Kati (approved):

PR #24, opened 2021-05-30
Code generators use symfony commands

Chemaclass:
## 📚 Description

Get rid of the current maker commands and implement the makers on top of symfony commands. This allows us to simplify a lot of the logic by having 2 simple commands:
- MakeFileCommand
- MakeModuleCommand

## 🔖 Changes

- Replace the "custom maker commands" in favor of the `symfony/console` commands.
- `FilenameSanitizer` uses `similar_text()` to guess the filename based on the percentage of similarity of the 4 different possibilities. 

### make:module examples

Screenshot 2021-05-30 at 11 03 08

### make:file examples

Screenshot 2021-05-30 at 11 08 55


src/CodeGenerator/Domain/FilenameSanitizer.php
```diff
@@ -0,0 +1,47 @@
...
+
+    public function sanitize(string $filename): string
+    {
+        $percents = [];
+        foreach (self::EXPECTED_FILENAMES as $expected) {
+            $percents[$expected] = similar_text($expected, $filename, $percent);
+        }
+
+        $maxVal = max($percents);
+        $maxValKeys = array_keys($percents, $maxVal, true);
```
Chemaclass, line 36: This is the magic! https://www.php.net/manual/en/function.similar-text.php
> similar_text — Calculate the similarity between two strings

Here I compare the similarity between the incoming filename and the potentially expected, and then the one with more percentage win 🥇

JesusValera (approved):

PR #26, opened 2021-06-06
Add long-name option to the commands

Chemaclass:
## 🔖 Changes

- Add `long-name` option to maker commands. 
  By default is false, but if you set it up it will add the module name to the generated files as a prefix for the classes.

Screenshot 2021-06-06 at 12 39 22


Tito-Kati (approved):
Nice 👌🏻

JesusValera (approved):
🔥

PR #28, opened 2021-06-08
Install phpstan level 5

Chemaclass:
## 🔖 Changes

- Added as require --dev the latest PHPStan with level 5


JesusValera (approved):
Nice one, let's look for the max level 👋🏼

PR #29, opened 2021-06-09
Fix new logo

Tito-Kati:
## 📚 Description

Seems that SVGs can also have bugs

## 🔖 Changes

- Created a unique path for each shape
- Changed the color to the correct one
- Removed some duplicate lines and others that should not appear


Chemaclass (approved):
Tuff guy 🤘 Niceeee 🥳

JesusValera (approved):
Cleaner ✨

PR #38, opened 2021-07-08
Support multiple config types

Chemaclass:
## 📚 Description

Currently, you can define only one type of config and in one single path pattern. For example:
```json
{
  "config": {
      "type": "php",
      "path": "config/*.php"
    }
}
```

## 🔖 Changes

- `gacela.json` allows an array of JSON with different configuration types. For example:
```json
{
  "config": [
    {
      "type": "env",
      "path": "config/.env*"
    },
    {
      "type": "php",
      "path": "config/*.php"
    }
  ]
}
```


JesusValera (approved):
Awesome, we can add Gacela to Laravel properly :))

PR #42, opened 2021-07-13
Add YamlConfigReader

JesusValera:
## 📚 Description

Add the ability to create `config` files with `YAML` or `YML` format.

Example file:
```yaml
config: 1
override: 5
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
     "require": {
         "ext-json": "*",
         "php": ">=7.4",
-        "symfony/console": "^5.2"
+        "symfony/console": "^5.3",
+        "symfony/yaml": "^5.3"
     },
     "require-dev": {
         "phpunit/phpunit": "^9.5",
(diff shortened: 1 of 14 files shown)


composer.json
```diff
@@ -17,7 +17,8 @@
     "require": {
         "ext-json": "*",
         "php": ">=7.4",
-        "symfony/console": "^5.2"
+        "symfony/console": "^5.3",
+        "symfony/yaml": "^5.3"
```
Chemaclass, line 21: I don't like the idea of introducing a new dependency on this project which doesn't really need to be required for most of the projects using Gacela... I like the idea of having this YamlConfigReader, but I would rather create another repository in the `gacela-project` org, so if someone needs this, they can explicitly require that other repo. But I wouldn't add this dependency to the main repo.

Chemaclass (changes requested):

JesusValera:
You are completely right @Chemaclass.
In order to avoid depend on new dependencies, I have created a new project 👍🏼 

https://github.com/gacela-project/gacela-yaml-config-reader

PR #43, opened 2021-07-14
Link the docs to the website

Chemaclass:
## 🔖 Changes

Remove the docs from the docs folder and point to the website in order to avoid duplicated work and unify the docs source of truth


JesusValera (approved):
Clever!

PR #47, opened 2021-07-29
Allow optional config readers in gacela.json

Chemaclass:
## 📚  Context

If a "config reader item" is specified in `gacela.json` it must be set up in the script as, for example:
```php
/*
 * This is an example of how can you create your own config reader.
 * The key 'custom' in the ConfigReaders will parse the files found
 * by the gacela.json config:
 * {
 *   "type": "custom",
 *   "path": "config/*.custom"
 * }
 */
Config::setConfigReaders([
    'php' => new PhpConfigReader(),
    'custom' => new CustomConfigReader(),
]);
```

But, what happens if you have that "custom config reader item" BUT you don't set up using `Config::setConfigReaders()`? It will simply throw a `ConfigReaderException` because the config files were fetched from `config/*.custom` but there is no ConfigReader assigned to read that file. 

## 💡 Goal

Make it possible to define optional "config reader items" so in case there is no ConfigReader found for a file, it will simply skip them and move to the next config file without throwing an error.

## 🔖 Changes

-  Throw a `ConfigReaderException::notSupported()` if the config reader item is non optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Non released yet
+
+- Allow optional config readers in gacela.json
+
 ### 0.6.0
 ##### 2021-07-27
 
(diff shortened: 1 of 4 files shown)


Chemaclass:
Wrong idea. I will rework it entirely to make it simpler

PR #49, opened 2021-07-31
Fix bug in GlobalKey for Modules with multi names

Chemaclass:
## 🤔 Context

The `GlobalKey::fromClassName()` wasn't working as expected for modules with more than one word. 

## 🔖 Changes

- Fix the bug that should unify all keys for the class resolver.


JesusValera (approved):
👏🏼 🦙

PR #51, opened 2021-08-02
Class resolver with dependencies

Chemaclass:
## Background

Currently, it's not possible to inject dependencies to the Factory, so when the Facade calls its Factory, there is no way to inject dependencies on it. This is a bit unfortunate because this way we cannot get, for example, the entity repositories from Symfony or other frameworks, in order to inject them while constructing your modules domain services. 

### Add auto-wire to your Gacela Factories. 

The constructor dependencies from your Factories will be resolved automagically if they are concrete classes. And if they are interfaces (which is the recommended thing; depend on abstractions (interfaces) and not concretions), then you can simply define which concrete class should the auto-wire instantiate when it finds that interface.

An example:

```php
# gacela.php

return [
    'config' => [/* ... */],
    'dependencies' => [
        YourInterface::class => TheConcreteClass::class,
    ],
];
```

```php
final class Factory extends AbstractFactory
{
    private YourInterface $foo;

    public function __construct(YourInterface $foo) 
    {
        $this->foo = $foo;
    }

    public function createYourService(): YourService
    {
        return new YourService($this->foo);
    }
}
```

## Callables

If instead of defining what concrete class does you want to create when there is a particular interface, you want to inject/use a particular and existing object, you can do that by using a callable function. This is useful, for example, to define the ManagerRegistry interface from Symfony

For example:

```php
# gacela.php

/** @var Kernel $kernel */
$kernel = GlobalServices::get('symfomy/kernel');

return [
    'config' => [
        "type" => "env",
        "path" => ".env*",
        "path_local" => ".env",
    ],
    'dependencies' => [
        ManagerRegistry::class => static fn() => $kernel
            ->getContainer()
            ->get('doctrine'),
    ],
];
```


JesusValera (approved):
We need to keep working on the autowiring, but for now it's nice 🤙🏻

PR #52, opened 2021-08-04
Replace config file from json to php

JesusValera:
## 📚 Description

The idea is to replace the Gacela config file from `gacela.json` to a plain PHP file `gacela.php`.
This way, the code is more modulable and fits better in the project for future refactorizations.


src/Framework/Config/GacelaFileConfig/GacelaPhpConfig.php
```diff
@@ -0,0 +1,75 @@
...
+            $c = GacelaPhpConfigItem::fromArray($config);
+            $result[$c->type()] = $c;
+        }
+
+        return $result;
+    }
+
+    private static function isMultiConfigFile(array $config): bool
+    {
+        return count(array_filter($config, 'is_array')) >= 1;
```
Chemaclass, line 58: What is this? lol 😂  
I think I know what you're trying to do... but maybe we can have something more readable 🤔

Chemaclass (approved):

PR #53, opened 2021-08-08
Refactor CodeGenerator module

Chemaclass:
## 🔖 Changes

- Use the `Facade` (and not the `Factory`) in the `./gacela script`.
- Move existing interfaces from Infrastructure to Domain layers.
- Create interfaces for the different Domain services
- Restructure and group Domain services by their responsibilities 
- Move the `TEMPLATE_BY_FILENAME_MAP` away from the Domain to the `DependencyProvider`, so it can be fetched from the Factory and injected into the Domain service which needs this logic (idea +- from [Emmanuel Valverde](https://twitter.com/evrtrabajo/status/1424013798885797890))
- Remove unnecesary `CodeTemplateInterface`


JesusValera (approved):
🧠 ⚡

PR #55, opened 2021-08-10
Some edge-cases in DependencyResolver

JesusValera:
## 📚 Description

The `DependencyResolver` class now is smarter.
Add a lot of tests to check the resolver works as expected (_probably we should work on the class-naming in a future iteration_).


Chemaclass (approved):
Wow, nice clean-up! 🧼

PR #58, opened 2021-08-12
Refactor tests related to binding interfaces

JesusValera:
## 📚 Description

Split the old `UsingConfigInterfacesMapping` into `BindingInterfacesInGacelaConfigFile` & `BindingInterfacesWithInnerDependencies`.

The first one (`BindingInterfacesInGacelaConfigFile`) check it is possible to bind classes as follows:
```php
// Resolve an abstract class from a concrete instance of a class
AbstractClass::class => new SpecificClass(true, 'string', 1, 1.2, ['array']),

// Resolve anonymous-classes/callables from abstract classes and interfaces
AbstractFromAnonymousClass::class  => $this->resolveAbstractFromAnonymousClass(),
AbstractFromCallable::class => $this->resolveAbstractFromCallable(),
InterfaceFromAnonymousClass::class => $this->resolveInterfaceFromAnonymousClass(),
InterfaceFromCallable::class => $this->resolveInterfaceFromCallable(),
```
Those methods are just functions that return anonymous classes or functions with anonymous classes like:
```php
return new class() implements/extends ...
# OR
return static fn () => new class() implements/extends ...
```
---
The second one (`BindingInterfacesWithInnerDependencies`) check it is possible to bind classes as follows AND if the `$globalServices` works properly:
```php
# init.php
Gacela::bootstrap(['env' => 'prod']);

# gacela.php
GreeterGeneratorInterface::class => IncorrectCompanyGenerator::class // Both as `::class`

// And the value in `$globalServices´ exists
$exists = false;
if ('prod' === $globalServices['env']) {
    $exists = true;
}

# $exists is TRUE.
```


tests/Integration/Framework/BindingInterfacesInGacelaConfigFile/IntegrationTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+declare(strict_types=1);
+
+namespace GacelaTest\Integration\Framework\BindingInterfacesInGacelaConfigFile;
+
+use Gacela\Framework\Gacela;
+use GacelaTest\Integration\Framework\BindingInterfacesInGacelaConfigFile\LocalConfig\Domain\AbstractFromAnonymousClass;
+use GacelaTest\Integration\Framework\BindingInterfacesInGacelaConfigFile\LocalConfig\Domain\AbstractFromCallable;
+use GacelaTest\Integration\Framework\BindingInterfacesInGacelaConfigFile\LocalConfig\Domain\InterfaceFromAnonymousClass;
+use GacelaTest\Integration\Framework\BindingInterfacesInGacelaConfigFile\LocalConfig\Domain\InterfaceFromCallable;
```
Chemaclass, line 11: Could we come up with a better naming for this module name instead of `LocalConfig`?

Chemaclass (commented):
I really like all these assertions in `BindingInterfacesInGacelaConfigFile` and refactoring. But I got one comment about the naming of the module for the integration test that we are using.

tests/Integration/Framework/BindingInterfacesWithInnerDependencies/gacela.php
```diff
@@ -0,0 +1,31 @@
...
+
+declare(strict_types=1);
+
+use Gacela\Framework\AbstractConfigGacela;
+use GacelaTest\Integration\Framework\BindingInterfacesWithInnerDependencies\LocalConfig\Domain\GreeterGeneratorInterface;
+use GacelaTest\Integration\Framework\BindingInterfacesWithInnerDependencies\LocalConfig\Infrastructure\CorrectCompanyGenerator;
+use GacelaTest\Integration\Framework\BindingInterfacesWithInnerDependencies\LocalConfig\Infrastructure\IncorrectCompanyGenerator;
+
+/**
+ * This integration-test does two things:.
```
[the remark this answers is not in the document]
Chemaclass: This is because of the phpcs rule `phpdoc_summary`. I will change it to don't force a dot in the end always.

tests/Integration/Framework/BindingInterfacesInGacelaConfigFile/IntegrationTest.php
Chemaclass: As we agreed, we'll create a follow-up PR renaming this directory, because it's not the only one but some other integration tests have the same "copy/paste" problem, so we'll refactor them all at once because they are all related due to the same reason.

Chemaclass (approved):

PR #62, opened 2021-10-03
Config from global services

Chemaclass:
## 📚 Description

@jenshaase gave me the idea of allowing the user to specify a custom config without the need of the `gacela.php` file itself by specifying the 'config' somehow when bootstrapping Gacela -> https://github.com/phel-lang/phel-scaffolding/pull/14#issuecomment-932802547

## 🔖 Changes

- This is a backward-compatible addition. The `gacela.php` file still works as before
- The addition is that now you can also specify the config of your project (where do you have your config files and their types) as for example:
```php 
Gacela::bootstrap($projectRootDir, [
    'config' => [
        'type' => 'php',
        'path' => 'phel-config.php',
        'path_local' => 'phel-config-local.php',
    ],
]);
``` 

## 💡 Next steps

Maybe we don't need the `gacela.php` file anymore? Should we deprecate it? it is still useful? 
Because right now you can simply pass the 'config' and the 'mapping-interfaces' as global services when doing `Gacela::bootstrap()`, maybe there is no need for the `gacela.php` file 🤔 
Or maybe it's actually good for the separation of concerns. 

For now, I think this is good to have such flexibility in the project. Any ideas (pros / const) related to this topic are welcome 🙏


JesusValera (approved):
Nice, but I still like the idea of allowing the user to define a ```gacela.php``` file independently.
Also, I'm not sure but maybe it's not possible in Laravel/Symfony framework on this way, I should check it out first 🤔

PR #63, opened 2021-10-31
Improved type annotation in AbstractClassResolver and Locator

jenshaase:
## 📚 Description

Improves the type annotation in `AbstractClassResolver` and `Locator` to have better psalm support. Additionally the method `Locator::newInstance` will now return `null` instead of a string if the class doesn't exists.

## 🔖 Changes

- Added type annotation in `AbstractClassResolver`
- Added type annotation in `Locator`
- `Locator::newInstance` returns `null` if the class doesn't exist.

### DIFF
--- a/src/Framework/ClassResolver/AbstractClassResolver.php
+++ b/src/Framework/ClassResolver/AbstractClassResolver.php
@@ -79,15 +79,24 @@ abstract class AbstractClassResolver
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Chemaclass (approved):
Wow! I love it, great idea 💡

JesusValera (approved):

PR #66, opened 2021-12-31
Remove deprecated gacela config as array

Chemaclass:
### 📚 Description

Gacela config as array was deprecated since a few months, so I decided to removed the deprecated feature.
**BEFORE** this was possible (not anymore!):
```php
<?php # gacela.php
return [
   'config' => [
        // ... config setup
    ], 
   'mapping-interfaces' => [
        // ... interfaces setup
    ],
];
```

### 🔖 Changes

- It's not possible to return an array from `gacela.php`

You have to return an anon-function that returns a class that extends from `AbstractConfigGacela`. For example:
```php
<?php # gacela.php
use Gacela\Framework\AbstractConfigGacela;

return static function () {
    return new class() extends AbstractConfigGacela {
        public function config(): array
        {
            return [
                // ... config setup
            ];
        }
        public function mappingInterfaces(): array
        {
             return [
                // ... interfaces setup
            ];
        }
    };
};
```

### 💡 Why?

Because the class provides you a wrapper with the methods that you can override (simplifying the data mapping from-to gacela's config), and it gives you access to the **GlobalServices** feature.
See more in: https://gacela-project.com/about-gacela/config/


JesusValera (approved):
Thanks 😊👍🏼

PR #67, opened 2022-01-01
Ensure that Gacela is compatible with PHP 8.1

Chemaclass:
## 🔖 Changes

- Add composer cache on GitHub CI.
- Add PHP 8.1 to the `matrix.php` when running the tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,16 +11,29 @@ jobs:
   coding-guidelines:
     name: Coding Guidelines
     runs-on: ubuntu-latest
-
+    strategy:
+      fail-fast: false
+      matrix:
+        php: ['7.4']
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
+      - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -11,16 +11,29 @@ jobs:
   coding-guidelines:
     name: Coding Guidelines
     runs-on: ubuntu-latest
-
+    strategy:
+      fail-fast: false
+      matrix:
+        php: ['7.4']
```
JesusValera, line 17: Only 7.4 here?

Chemaclass: Yes, because it's to run the coding-guidelines (php-cs-fixer), I don't think we have to run it for all php versions.

Chemaclass: I removed the matrix strategy for non tests jobs.

JesusValera: It makes sense 👌🏼👌🏼

JesusValera (approved):

PR #72, opened 2022-01-15
Global services as param on mapping interfaces

Chemaclass:
## 📚 Description

The GlobalServices are injected directly into `mappingInterfaces()` and not via the constructor.

### BEFORE

In order to use the global services, you had to pass them into the constructor such as this:
```php
return static fn (array $globalServices) =>  new class($globalServices) extends AbstractConfigGacela {
    public function mappingInterfaces(): array
    {
        if ('yes!' === $this->getGlobalService('isWorking?')) {
            $interfaces[GreeterGeneratorInterface::class] = CorrectCompanyGenerator::class;
        }

        return $interfaces;
    }
};
```
so you could use the `getGlobalService()` method.

### AFTER

The GlobalServices are injected directly into the `mappingInterfaces` method. The code above is obsolete. 
Now it will look like:
```php
return static fn () => new class () extends AbstractConfigGacela {
    public function mappingInterfaces(array $globalServices): array
    {
        if ('yes!' === $globalServices['isWorking?']) {
            $interfaces[GreeterGeneratorInterface::class] = CorrectCompanyGenerator::class;
        }

        return $interfaces;
    }
};
```

This is easier to use, and you don't need to care about the `AbstractConfigGacela` constructor at all. You just need to follow the signature of the parent method.


JesusValera (approved):
Nice catch, now it is much better 👍🏼

src/Framework/AbstractConfigGacela.php
```diff
@@ -34,9 +15,11 @@ public function config(): array
...
 
     /**
+     * @param array<string,mixed> $globalServices
+     *
      * @return array<string,string|callable>
      */
-    public function mappingInterfaces(): array
+    public function mappingInterfaces(array $globalServices): array
     {
         return [];
```
JesusValera, line 24:
```suggestion
    public function mappingInterfaces(array $globalServices): array
    {
        return $globalServices;
```

What do you think @Chemaclass?

Chemaclass: Good eye, but it's not what we want. The global-services are *not* the same as the mapping-interfaces. I updated the signature code to reflect this intention clearer.
```php
/**
 * @param array<string,mixed> $globalServices
 *
 * @return array<class-string,class-string|callable>
 */
public function mappingInterfaces(array $globalServices): array
{
    return [];
}
```

JesusValera: That's true.

PR #79, opened 2022-02-06
Move config-readers as globalServices and gacela.php

Chemaclass:
## 📚 Description

We define custom configReaders as 3rd param in `Gacela::bootstrap()`. There are some disadvantages on this:
- More arguments for the `bootstrap()` increase the complexity of the main entry point of Gacela
- The configReaders are not possible to be defined in `gacela.php`
For these reasons, for example, I thought it will be better treating the `configReaders` the same as the existing `config` or `mappingInterfaces` concepts, which you are able to alter in your `gacela.php`, and also inside the `globalServices`.

## 🔖 Changes

- You can define in your `gacela.php` the `configReaders()` method
- You can define the `config-readers` as key in the `$globalServices`
- Change the bootstrap signature to this:

```php
final class Gacela
{
    /**
     * Define the entry point of Gacela.
     *
     * @param array<string,mixed> $globalServices
     */
    public static function bootstrap(string $appRootDir, array $globalServices = []): void
    {/* ... */}
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 - Add getAppRootDir() to AbstractConfig.
 - Deprecate getApplicationRootDir() from Config. Use getAppRootDir() instead.
-- The Gacela::bootstrap() now accepts a 3rd parameter to add config readers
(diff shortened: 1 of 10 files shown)


JesusValera (approved):
👏🏼

PR #81, opened 2022-02-07
Allow using config from custom env

Chemaclass:
## 📚 Description

Currently there is no possibility of defining some config for certain specific environment. So you end-up having the same config in all possible environments.

## 🔖 Changes

Added `APP_ENV` environment key, where you can define a value that you want, and that value will be added as suffix for your config file patterns.

#### For example

When `APP_ENV=env` then this is order of reading the config files, so the latest overrides the previous key-values:

- default.php
- default-dev.php
- local.php (defined in the gacela.php under path_local key)

> Check the UsingConfigFromCustomEnv Integration tests.

### 🤔 Follow up

Expose the possibility of creating custom `AbsolutePathStrategyInterface` via the globalInterfaces (when bootstrapping), and so, it would be interesting to be able to assign a particular path strategy to a concrete config-file-extension somehow.

### DIFF
--- a/src/Framework/Config/ConfigLoader.php
+++ b/src/Framework/Config/ConfigLoader.php
@@ -14,14 +14,18 @@ final class ConfigLoader
 
     private PathFinderInterface $pathFinder;
 
+    private string $configFileNameSuffix;
+
     public function __construct(
         string $applicationRootDir,
         GacelaConfigFileFactoryInterface $configFactory,
-        PathFinderInterface $pathFinder
+        PathFinderInterface $pathFinder,
+        string $configFileNameSuffix = ''
     ) {
         $this->applicationRootDir = $applicationRootDir;
         $this->configFactory = $configFactory;
         $this->pathFinder = $pathFinder;
+        $this->configFileNameSuffix = $configFileNameSuffix;
     }
 
     /**
@@ -109,10 +113,29 @@ final class ConfigLoader
 
     private function generateAbsolutePath(string $relativePath): string
     {
+        // place the file suffix right before the file extension
+        $dotPos = strpos($relativePath, '.');
+        $suffix = $this->getConfigFileNameSuffix();
+
+        if ($dotPos !== false && !empty($suffix)) {
+            $relativePathWithFileSuffix = substr($relativePath, 0, $dotPos)
+                . '-' . $this->getConfigFileNameSuffix()
+                . substr($relativePath, $dotPos);
+        } elseif (!empty($suffix)) {
+            $relativePathWithFileSuffix = $relativePath . $this->getConfigFileNameSuffix();
+        } else {
+            $relativePathWithFileSuffix = $relativePath;
+        }
+
         return sprintf(
             '%s/%s',
             $this->applicationRootDir,
-            $relativePath
+            $relativePathWithFileSuffix
         );
+    }
+
+    private function getConfigFileNameSuffix(): string
+    {
+        return $this->configFileNameSuffix;
     }
 }
--- a/src/Framework/Config/ConfigLoader.php
+++ b/src/Framework/Config/ConfigLoader.php
@@ -54,7 +54,7 @@ final class ConfigLoader
     {
         $configGroup = [];
         foreach ($gacelaFileConfig->getConfigItems() as $configItem) {
-            $absolutePath = $this->generateAbsolutePath($configItem->path());
+            $absolutePath = $this->generateAbsolutePathWithSuffix($configItem->path());
             $matchingPattern = $this->pathFinder->matchingPattern($absolutePath);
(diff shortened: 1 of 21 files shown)


src/Framework/Config/ConfigLoader.php
```diff
@@ -107,6 +111,25 @@ private function readLocalConfigFile(GacelaConfigFile $gacelaConfigFile): array
         return array_merge(...array_filter($result));
     }
 
+    private function generateAbsolutePathWithSuffix(string $relativePath): string
+    {
+        // place the file suffix right before the file extension
```
JesusValera, line 116: What about `.env` files? Did you already take them into account? 😄

JesusValera (commented):
So, the key from `APPLICATION_ENV` should match with the config file ending -> `default-dev.php` => `dev`.

Is that right? :D

src/Framework/Config/ConfigLoader.php
JesusValera: Also, the order (when `APPLICATION_ENV=env`) is the following?

1. `default.php`
2. `default-dev.php`
3. defined in the `gacela.php` under `path_local` key, e.g: `local.php`

Chemaclass: Actually, that's the desired behaviour, yes! But I think it doesn't work right now like that... I need to refactor and extract this functionality `generateAbsolutePathWithSuffix` on its own class, so I can rite unit tests for each behaviour. Thanks for spotting this.

Also, I had an idea of allowing the user to define different config-name-with-env merging strategies that will make this feature totally customisable and extensible. This will naturally come afterwards.

Chemaclass:
I think you got confused because the code wasn't clear enough, my bad. There are two concepts which you mixed: 
1) the APP_ENV, which defines the current environment where the app is running, 
2) the `.env` config files, which are simple key-values config files.

The important idea behind this PR is that we need to allow the user to specify config files which have priority over other config files based on the server environment. This way, you can deploy your app in qa `default-qa.php`, stage `default-stage.php`, prod `default-prod.php`, and have different config key-values for each environment.

In this case, the local config file `local.php` will be ignored by git, so there is no possibility of source control (it’s aimed just for local dev/testing purposes), although your idea of using it to have that local config on each environment is not bad, there are some disadvantages, especially when considering feature-flags (where you need to deploy that feature in prod and immediately after go to the server and add the flag to `false` or similar...), with this PR we just need to commit the feature flag as false in the `default-prod.php`, and that's all.

That said, of course there are different types of config files that we want to allow, and also different strategies for the "config-file-name-environment", and this is where the real complexity is.

Right now it's working perfectly for php config files. I still need to think and prepare the strategy for the `.env` files such as:
```
# In all environments, the following files are loaded if they exist,
# the latter taking precedence over the former:
#
#  * .env                contains default values for the environment variables needed by the app
#  * .env.$APP_ENV       committed environment-specific defaults
#
# Real environment variables win over .env files.
```

JesusValera:
@Chemaclass ok, I see.
Yes, I mess up and though the method `normalizePathPatternWithEnv()` was an "exception" for the `.env` files. Makes sense now 👍🏼 

How does the project know which environment should be used?
Maybe we can pass to the `Gacela::bootstrap()` a third argument with the current environment?

Chemaclass:
I don't like the idea of inflate the bootstrap method adding more arguments to it... For now I think it would be enough to simply put it as a global env variable. See `tests/Integration/Framework/UsingConfigFromCustomEnv/IntegrationTest.php`
```php
putenv('APP_ENV=dev');
Gacela::bootstrap(__DIR__);
```
Anyway, I am still thinking about this looking for the best approach. I still don't know, I am on "discovering process" mode 😛

JesusValera:
I see.
In Symfony/Laravel it is not a problem because the bootstrap files (`bin/console` or `artisan` & `public/index.php`) files, the very first thing they do is read the `DotEnv()` variables.
But in the case we are using no framework at all, we are "coupled" to define manually somewhere (in the `public/index.php` maybe?) the env we want to load as you did.
And if we do that, the env will be the same always, it doesn't matter if we deploy the app on a server because the `index.php` will have the `APP_ENV` variable already set, do you get what I mean @Chemaclass?

Chemaclass:
It's not a matter of frameworks (symfony/laravel) but of environment variables. 
It think it's quite normal to define the app_env as env variables; for example you can do this easily with docker:
```yaml
version: '3.7'
services:
  your-app-local:
    environment:
      APP_ENV: dev
     # etc

  your-app-prod:
    environment:
      APP_ENV: prod
     # etc
```
> This is a silly example, but it's useful to show the point.

JesusValera:
Ok, I got it 👍🏼

JesusValera (approved):

PR #84, opened 2022-02-12
Create CustomServices

Chemaclass:
## 📚 Description

Gacela has currently only the concepts of Facade, Factory, Config and DependencyProvider (aka: native Gacela services), but what about allowing defining custom services that are accesible from the Facade or Factory, and they have access to all those Gacela native services? 
For example, being able to create a Repository in your modules under the Infrastructure namespace.

## 🔖 Changes

- Introduce "flexible services" concept
  - You can define them in your `gacela.php` or directly in the `Gacela::bootstrap()`
- The Facade and Factory have access to the "flexible services" out the box
  - you simply have to specify the paths and resolvable-types that you want to be able to instantiate.

For example:
```php
Gacela::bootstrap(__DIR__, [
    'custom-service-paths' => ['Application', 'Infrastructure'],
]);
```
It will search for CustomServices inside the namespace "Application" and "Infrastructure" (inside the current module). The `resolvableType` will be the method name used without the prefix "get".

For example, you the Repository from the Infrastructure directory/namespace inside a module, can be called using `getRepository()` from the Facade or Factory. You can use PHPDocs for help and autocompletion like: 
```php
namespace App\CusomModule;

/**
 * @method Repository getRepository()
 */
final class Facade extends AbstractFacade
{/*...*/}
```
and then the Repository itself, which has access to the Factory and Config from the same namespace/layer or the module:
```php
namespace App\CusomModule\Infrastructure;

use App\FlexibleApiModule\Config;
use Gacela\Framework\AbstractCustomService;

/**
 * @method Factory getFactory()
 * @method Config getConfig()
 */
final class Repository extends AbstractFlexibleService
{/*...*/}
```

⭐️ This is really powerful, because you can define inside that "Infrastructure" layer/directory, a specific Factory/Config for that context, and in case there is none, it will move up and take the one from the module itself.


JesusValera (approved):
Such a good idea 💡

PR #91, opened 2022-02-14
Custom services must implement CustomServiceInterface

Chemaclass:
## 📚 Description

Currently the custom services must extend from AbstractCustomService, but it would be nice to don't force inheritance when it's not required. Although, with the AbstractCustomService you can use the getFactory() and getConfig() of the module, your custom services might don't need that, so the interface might be sufficient.

## 🔖 Changes

- Change from forcing extending `AbstractCustomService` to just implement `CustomServiceInterface`.

The advantage of using custom services is that they are accesible from gacela services AND they have an autoloader for their dependencies. The mapping-interfaces config is also available for the custom services too.

> 💡 Kudos to @JesusValera who gave me the idea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### Unreleased
+
+- Added `CustomServiceInterface` that custom services must implement.
+  - Although, the `AbstractCustomService` implements that interface.
+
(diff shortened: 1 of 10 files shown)


JesusValera (approved):
😄

PR #92, opened 2022-02-14
Add cachedClassNames for ClassNameFinder

Chemaclass:
## 📚 Description

Currently the custom services must start with get prefix.

## 🔖 Changes

- Allow calling custom services with `get` prefix or without prefix at all.
- Added `$cachedClassNames` so there is no need to resolve the same classInfo if it was already found it earlier.
- Added feature tests to ensure the recursive class locator from inner namespace level from custom services to upper namespace level. So a custom service can access the Config from the module, for example.


src/Framework/CustomServiceAwareTrait.php
```diff
@@ -13,7 +13,7 @@ trait CustomServiceAwareTrait
 
     public function __call(string $name, array $arguments = []): CustomServiceInterface
     {
-        $resolvableType = ltrim($name, 'get');
+        $resolvableType = ucfirst(ltrim($name, 'get'));
```
JesusValera, line 16: Don't forget this method is not working properly, we should use a regex_replace like \^get\ to empty?

tests/Feature/Framework/CustomServicesLevel/CustomModule/Level1/Level2/Level3/HappyService3.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace GacelaTest\Feature\Framework\CustomServicesLevel\CustomModule\Level1\Level2\Level3;
+
+use Gacela\Framework\CustomServiceInterface;
+
+final class HappyService3 implements CustomServiceInterface
+{
+    public function greet(string $name): string
+    {
+        return "Hi, $name!";
```
JesusValera, line 13: I would put maybe something like: "Hi from level #, $name", what do you think?
This way is more clear what we are doing.

JesusValera (commented):
Some adjustments.

JesusValera (approved):

PR #93, opened 2022-02-16
Create basic AbstractTransfer

Chemaclass:
## 📚 Description

I was thinking about adding a complete new feature to Gacela like an abstract class which facilitates the creation of DTOs.

## 🔖 Changes

- Added `AbstractTransfer` class.
  - The suffix `Transfer` is from "Data Transfer Object", not from "Transportation" or similar.

I was thinking about using `AbstractDto`, or `AbstractDTO`, or `AbstractTransfer`...

## 🤔 Follow up ideas

As follow up, we could think about creating a Gacela command, where you specify a file which is suppose to extend the `AbstractTransfer`, and then alter it to add the PHPDoc above the class name to add all getters and setters. 

We could even pass a directory/ies and search in all those possible classes of interest.

### DIFF
--- a/tests/Unit/Framework/Transfer/AbstractTransferTest.php
+++ b/tests/Unit/Framework/Transfer/AbstractTransferTest.php
@@ -52,14 +52,14 @@ final class AbstractTransferTest extends TestCase
 
     public function test_set_wrong_type_from_setter(): void
     {
-        $this->expectErrorMessageMatches('/.*Typed property .*PersonTransfer.*age must be int or null, float used/');
+        $this->expectErrorMessageMatches('/.*PersonTransfer.*age.*/');
 
         (new PersonTransfer())->setAge(10.5);
     }
 
     public function test_set_wrong_type_from_array(): void
     {
-        $this->expectErrorMessageMatches('/.*Typed property .*PersonTransfer.*age must be int or null, float used/');
+        $this->expectErrorMessageMatches('/.*PersonTransfer.*age.*/');
 
         (new PersonTransfer())->fromArray(['age' => 10.5]);
     }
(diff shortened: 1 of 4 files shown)


tests/Unit/Framework/Transfer/AbstractTransferTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+namespace GacelaTest\Unit\Framework\Transfer;
+
+use Gacela\Framework\Transfer\UnknownPropertyException;
+use GacelaTest\Unit\Framework\Transfer\Fixtures\PersonTransfer;
+use PHPUnit\Framework\TestCase;
+
+final class AbstractTransferTest extends TestCase
+{
+    public function test_fluent_setters(): void
```
JesusValera, line 13: Both tests are doing the same (`test_fluent_setters()` + `test_getters()`), but one is setting the parameters fluently and the other not.
It doesn't make sense because both methods are `setting` & `getting` the information for the assertion.

I would suggest to change the test name to something like `test_properties_transfer_object_fluently` & `test_properties_object_transfer` maybe?

tests/Unit/Framework/Transfer/AbstractTransferTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+        (new PersonTransfer())->fromArray(['age' => 10.5]);
+    }
+
+    public function test_set_nonexistent_property_from_array(): void
+    {
+        $person = (new PersonTransfer())->fromArray(['non-existent' => 123]);
+
+        self::assertInstanceOf(PersonTransfer::class, $person);
+    }
```
JesusValera, line 73: What is this test doing?
I don't see any value, maybe we could check the parameters and assert the `non-existent` property is defined?

PR #97, opened 2022-02-20
Remove CustomService feature

Chemaclass:
## 📚 Description

After a lot of thinking and talking to @JesusValera, we agreed that this feature (introduced in the previous version) was adding almost no benefits and too much complexity. Using MappingInterfaces feature solves this problem is a more elegant way (and less magical), so we decided to drop this entire CustomService feature before is too late ⏰ 

> It was fun to try this, thought. And a lot of things were improved during the develop of this feature :)  

## 🔖 Changes

- Removed the entire CustomService feature.


JesusValera (approved):
🌚👌🏼🤙🏻

Issue #105, opened 2022-03-01
Recognize classes types by the abstract they extend rather than by name/prefix


Tito-Kati:
Gacela does a good job of helping to decouple the domain from everything else but is nonetheless coupled with the verbosity of the names of the classes that extend its abstracts.

It would be great if Gacela detects classes by the abstract they extend instead of by their name/prefix.

An example where this would be beneficial is if Gacela was used in a large project and suddenly the team decided to change the nomenclature of their "Gacela classes".

If the framework recognized the classes in this way, a progressive refactor strategy could be adopted instead of having to do all the work at once to apply the new names.

What do you think?

Chemaclass:
With this https://github.com/gacela-project/gacela/pull/106 now we'll be able to define multiple suffixes for gacela resolvable types. A follow up issue it could be to scan the module directory and look for the class that it's extending certain `...Abtract` class, and then pick the right one.

The biggest problem/complexity I see with this approach is the fact of scanning the directories and doing so much parsing in order to find the proper classes... This could be a big downside for the performance. For this reason, I wanted to first develop this concept (of explicitly allowing multiple suffixes that you have to know in advance; so you can write them down in gacela.php) and then the other thing could be postpone or even reconsider.

PR #106, opened 2022-03-02
Allow multiple overriding resolvable types

Chemaclass:
## 📚 Description

Related to the issue: https://github.com/gacela-project/gacela/issues/105 although, it might not be exactly the "requested feature itself", I think this is an easier implementation that can help us get there if needed.

This allows you to define not just one replacement for the resolvable types keywords (Factory, Config, DependencyProvider) but now you can define multiple keywords for the same resolvable type.

### For example
You can have modules with different suffixes for Factories, Config, and DependencyProvider gacela classes, like:
```php
[
    'Factory' => ['FactoryModuleA', 'FactoryModuleB', '...'],
    'Config' => ['ConfModuleA', 'ConfModuleB'],
    'DependencyProvider' => ['DepProModuleA', 'DepProModuleB'],
]
```

You could even define the default values in the begging and having as fallback other especial suffixes for certain modules:
```php
[
    'Factory' => ['Factory', 'FactoryModuleA', 'FactoryModuleB', '...'],
    'Config' => ['Config', 'ConfModuleA', 'ConfModuleB', '...'],
    'DependencyProvider' => ['DependencyProvider', 'DepProModuleA', 'DepProModuleB', '...'],
]
```

## 💡  Goal

Allow the developer to define different resolvable types, so they can have different naming/suffix on different modules.

## 🔖 Changes

- In the gacela.php the `overrideResolvableTypes()` allows a `list<string>|string`.
- Refactor `ClassInfo`. Added a named-constructor  `::fromObject`.

### DIFF
--- a/src/Framework/AbstractConfigGacela.php
+++ b/src/Framework/AbstractConfigGacela.php
@@ -70,14 +70,24 @@ abstract class AbstractConfigGacela
      * <code>
      * return [
      *     'Factory' => 'Creator',
-     *     'Config' => 'Configuration',
+     *     'Config' => 'Conf',
      *     'DependencyProvider' => 'Binding',
+     * ];
+     * // OR
+     * return [
+     *     'Factory' => ['Creator', '...'],
+     *     'Config' => ['Conf', '...'],
+     *     'DependencyProvider' => ['Binding', '...'],
      * ];
      * </code>
      *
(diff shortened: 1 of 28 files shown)


Tito-Kati (approved):
Nice MVP! Thanks you @Chemaclass!

JesusValera (approved):

PR #107, opened 2022-03-03
Make GacelaConfigFileFactory testable

Chemaclass:
## 📚 Description

This is pointing to `feature/allow-multiple-resolvable-type-names` because otherwise we will face unnecessary conflicts (due to changes on the same files). **We can merge this PR (into master) once the other one gets merged first.**

## 🔖 Changes

- Create a `FileIoInterface` so we can create unit tests for `GacelaConfigFileFactory`.

### DIFF
--- a/src/Framework/Config/FileIo.php
+++ b/src/Framework/Config/FileIo.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Gacela\Framework\Config;
+
+final class FileIo implements FileIoInterface
+{
+    public function existsFile(string $filePath): bool
+    {
+        return file_exists($filePath);
+    }
+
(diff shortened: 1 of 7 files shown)


src/Framework/Config/FileIo.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    public function existsFile(string $filePath): bool
+    {
+        return file_exists($filePath);
+    }
+
+    /**
+     * @return mixed
+     */
+    public function include(string $filePath)
```
JesusValera, line 17: I think maybe it would make more sense to split this method into a new interface?

I don't see a real "relationship" between the `file_exists` and the `include` 🤔

JesusValera (approved):

PR #109, opened 2022-03-08
Create ResolvableTypesConfig

Chemaclass:
## 📚 Description

Instead of returning an array, we were thinking (see the idea: https://github.com/gacela-project/gacela/discussions/100) about creating a class to encapsulate the interesting fields for each "gacela.php" method. The idea is to follow a similar approach for the other two methods: `config()` & `mappingInterfaces()`.

## 🔖 Changes

- Create `ResolvableTypesConfig` as argument for `overrideResolvableTypes()`.


JesusValera (approved):
🎉🎉

PR #110, opened 2022-03-08
Create MappingInterfacesResolver

Chemaclass:
## 📚 Description

Instead of defining the mapping interfaces as a raw array, we created a class to wrap the usage of that behaviour making it easier to use for the end-user with the parameter `MappingInterfacesResolver`

## 🔖 Changes

- Create a new class `MappingInterfacesResolver`, which will be the first argument received from the `mappingInterface()`


JesusValera (approved):
👌🏼👌🏼

PR #113, opened 2022-03-09
Use builder instead of resolve prefix

Chemaclass:
## 📚 Description

Rename using "Builder" instead of "Resolver" to make it easier to understand the pattern that it's being used here. Also, split the current `GacelaConfigFileFactory` into two classes, so it's waaaaay easier to write tests for all possible behaviours.

## 🔖 Changes

- Rename the new `ConfigResolver`,`MappingInterfacesResolver` and `SuffixTypesResolver` using "Builder" suffix.
- Split `GacelaConfigFileFactory` into two classes:
  - `GacelaConfigFromBootstrapFactory`
  - `GacelaConfigUsingGacelaPhpFileFactory`


src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactory.php
```diff
@@ -0,0 +1,57 @@
...
+use Gacela\Framework\Config\GacelaConfigBuilder\ConfigBuilder;
+use Gacela\Framework\Config\GacelaConfigBuilder\MappingInterfacesBuilder;
+use Gacela\Framework\Config\GacelaConfigBuilder\SuffixTypesBuilder;
+use Gacela\Framework\Config\GacelaConfigFileFactoryInterface;
+use Gacela\Framework\Config\GacelaFileConfig\GacelaConfigFile;
+
+final class GacelaConfigFromBootstrapFactory implements GacelaConfigFileFactoryInterface
+{
+    /** @var array<string,mixed> */
+    private array $globalServices;
```
JesusValera, line 16: Why not the PhpDoc here is?

```php
/**
 * @var array{
 *     config?: callable,
 *     mapping-interfaces?: callable,
 *     suffix-types?: callable,
 * } $configFromGlobalServices
 */
```

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactory.php
```diff
@@ -0,0 +1,57 @@
...
+         *     suffix-types?: callable,
+         * } $configFromGlobalServices
+         */
+        $configFromGlobalServices = $this->globalServices;
+
+        $configBuilder = new ConfigBuilder();
+        $configFromGlobalServicesFn = $configFromGlobalServices['config'] ?? null;
+        if ($configFromGlobalServicesFn !== null) {
+            $configFromGlobalServicesFn($configBuilder);
+        }
```
JesusValera, line 41: What about those lines in a private method?

The same with the mappingInterfaces and the suffixes 😄

tests/Unit/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactoryTest.php
```diff
@@ -0,0 +1,86 @@
...
+final class GacelaConfigFromBootstrapFactoryTest extends TestCase
+{
+    public function test_no_global_services_then_default(): void
+    {
+        $factory = new GacelaConfigFromBootstrapFactory([]);
+
+        self::assertEquals(GacelaConfigFile::withDefaults(), $factory->createGacelaFileConfig());
+    }
+
+    public function test_no_especial_global_services_then_default(): void
```
JesusValera, line 27:
```suggestion
    public function test_no_special_global_services_then_default(): void
```

tests/Unit/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactoryTest.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public function test_global_service_mapping_interfaces_with_global_services(): void
+    {
+        $factory = new GacelaConfigFromBootstrapFactory([
+            'mapping-interfaces' => function (
+                MappingInterfacesBuilder $interfacesBuilder,
+                array $globalServices
+            ): void {
+                assert($globalServices['globalServiceKey'] === 'globalServiceValue');
```
JesusValera, line 57: Is this assert required in the test?

Chemaclass: Good point, I will change it to:
```php
self::assertSame($globalServices['globalServiceKey'], 'globalServiceValue');
```
This is ensuring that the `$globalServices` contains the expected global key-values array .

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactory.php
Chemaclass: Because global services are a more abstract concept. This here is a way to let psalm/phpstan know the concrete types for these special keys that I am looking for. It's like a (soft)casting. It wouldn't be correct place it in this line 16.

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactory.php
Chemaclass: Interesting idea. I will try this something about it  👍🏼

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactory.php
```diff
@@ -25,13 +25,16 @@ public function __construct(array $globalServices)
...
-         * } $configFromGlobalServices
-         */
+        $configBuilder = $this->createConfigBuilder();
+        $mappingInterfacesBuilder = $this->createMappingInterfacesBuilder();
+        $suffixTypesBuilder = $this->createSuffixTypesBuilder();
+
+        return GacelaConfigFile::usingBuilders($configBuilder, $mappingInterfacesBuilder, $suffixTypesBuilder);
+    }
+
+    private function createConfigBuilder(): ConfigBuilder
```
JesusValera, line 35: What about? (same with others :) )

```php
    private function createConfigBuilder(): ConfigBuilder
    {
        /** @var callable|null $configFromGlobalServices */
        $configFromGlobalServices = $this->globalServices['config'] ?? null;

        $configBuilder = new ConfigBuilder();
        if ($configFromGlobalServices !== null) {
            return $configFromGlobalServices($configBuilder);
        }

        return $configBuilder;
    }
```

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigUsingGacelaPhpFileFactory.php
```diff
@@ -49,15 +49,34 @@ public function createGacelaFileConfig(): GacelaConfigFile
             throw new RuntimeException('Your anonymous class must extends AbstractConfigGacela');
         }
 
+        $configBuilder = $this->createConfigBuilder($configGacelaClass);
+        $mappingInterfacesBuilder = $this->createMappingInterfacesBuilder($configGacelaClass);
+        $suffixTypesBuilder = $this->createSuffixTypesBuilder($configGacelaClass);
+
+        return GacelaConfigFile::usingBuilders($configBuilder, $mappingInterfacesBuilder, $suffixTypesBuilder);
```
JesusValera, line 56: Just another alternative to think about, nor better or worse
```php
return GacelaConfigFile::usingBuilders(
    $this->createConfigBuilder($configGacelaClass),
    $this->createMappingInterfacesBuilder($configGacelaClass),
    $this->createSuffixTypesBuilder($configGacelaClass)
);
```

JesusValera (approved):

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigUsingGacelaPhpFileFactory.php
Chemaclass: I thought about this too, but my first impression is that it's "too much compress"... maybe we'll change our mind and refactor it in the future, but for now I would prefer keep it with the variables :)

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactory.php
Chemaclass: Interesting, but it wouldn't work because of `return $configFromGlobalServices($configBuilder)`. Thanks for the idea, anyway 💡

Issue #115, opened 2022-03-11
Make public getAppRootDir()


Chemaclass:
## Current behaviour

Currently the `AbstractConfig->getAppRootDir()` is protected. We want to make that method public.
- Where? 👉 `Gacela\Framework\AbstractConfig`

### Why?

The idea behind this method is to easily access the root dir of the project from the Factory, but as protected, you have to override the same method in your Config module, something that doesn't make sense.

```php
final class YourCustomConfig
{
    public function getAppRootDir(): string # override
    {
        return parent::getAppRootDir();
    }
}
```

This wont be necessary at all if the `getAppRootDir()` is already public from the `AbstractConfig` class.

Chemaclass:
Done https://github.com/gacela-project/gacela/pull/116

PR #116, opened 2022-03-11
Make public getAppRootDir() #115

webcraftniray:
## 📚 Description

Allow easy access the root dir of the project from the Factory.

## 🔖 Changes

Changes getAppRootDir() method to 'public'

### DIFF
--- a/src/Framework/AbstractConfig.php
+++ b/src/Framework/AbstractConfig.php
@@ -20,12 +20,9 @@ abstract class AbstractConfig
         return Config::getInstance()->get($key, $default);
     }
 
-	/**
-	 *
-	 * Allow easy access the root dir of the project from the Factory
-	 *
-	 * @return string
-	 */
+    /**
+     * Allow easy access to the root dir of the project.
+     */
 
     public function getAppRootDir(): string
     {
--- a/src/Framework/AbstractConfig.php
+++ b/src/Framework/AbstractConfig.php
@@ -20,7 +20,14 @@ abstract class AbstractConfig
         return Config::getInstance()->get($key, $default);
     }
 
-    protected function getAppRootDir(): string
+	/**
+	 *
+	 * Allow easy access the root dir of the project from the Factory
+	 *
+	 * @return string
+	 */
+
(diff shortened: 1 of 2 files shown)


src/Framework/AbstractConfig.php
```diff
@@ -20,7 +20,14 @@ protected function get(string $key, $default = Config::DEFAULT_CONFIG_VALUE)
         return Config::getInstance()->get($key, $default);
     }
 
-    protected function getAppRootDir(): string
+	/**
+	 *
+	 * Allow easy access the root dir of the project from the Factory
+	 *
+	 * @return string
```
Chemaclass, line 27: If the types are already at code level (for argument types or returning types) we don't need/want to duplicate and define the return type again as PHPDoc. We use PHPDoc types only for array/list, generics, or when we want to provide a description to it.

src/Framework/AbstractConfig.php
```diff
@@ -20,7 +20,14 @@ protected function get(string $key, $default = Config::DEFAULT_CONFIG_VALUE)
         return Config::getInstance()->get($key, $default);
     }
 
-    protected function getAppRootDir(): string
+	/**
+	 *
+	 * Allow easy access the root dir of the project from the Factory
```
Chemaclass, line 25: Maybe something like this would be better, what do you think?
```suggestion
	 * Allow easy access to the root dir of the project.
```

src/Framework/AbstractConfig.php
```diff
@@ -20,7 +20,11 @@ protected function get(string $key, $default = Config::DEFAULT_CONFIG_VALUE)
         return Config::getInstance()->get($key, $default);
     }
 
-    protected function getAppRootDir(): string
+    /**
+     * Allow easy access to the root dir of the project.
+     */
+
```
Chemaclass, line 26: Can you remove this empty line between the PHPDoc and the method, please? 🙏

Chemaclass (approved):
Nice! Thanks! 👌🏼

PR #120, opened 2022-03-21
Gacela cache

Chemaclass:
## 📚 Description

Allow creating a "ResolvableClassNamesCache" file (stored as a json) with the mapping between the classInfoKey and its ResolvableTypeClassName, so there is no need to loop over and find the class-name if it is already cached.

## 🔖 Changes

- Implemented the MVP for the "ResolvableClassNamesCache" idea.


JesusValera (commented):
The PR seems nice, but as we already discuss, we should think if this PR really makes sense (at least for now).

Maybe we are trying to reduce the bootstrapping time, but in fact we are doing the opposite  because the `file_get_contents` function.

PR #123, opened 2022-03-24
Encapsulate global services inside setup

Chemaclass:
## 📚 Description

Currently, the concept of "global services" is mixing the gacela setup keys like `config`, `mapping-interfaces`, and `suffix-types` with the actual "global services" you might have (especially for the `mapping-intefaces` context (as 2 arg)).

For this reason, I wanted to encapsulate the real global-services you might want to define inside its own key. And rename that 2nd argument from `Gacela::bootstrap()` to `$setup`.

## 🔖 Changes

- Rename the current concept of "global services" to "setup".
- Create a new key inside the "setup" array named "global services". 
- Pass this new "global services" array as 2nd arg to the "mapping interfaces" method.


Chemaclass:
I didn't use "configuration" because it could be misleading with the "config" key, don't you think, @JesusValera? 🤔

JesusValera:
> I didn't use "configuration" because it could be misleading with the "config" key, don't you think, @JesusValera? Anyway, we could rename this in another iteration if we find a better name.

Yes, that's right.
Ok, let's go for `setup` instead :)

JesusValera (approved):

PR #127, opened 2022-03-25
Create SetupGacela to replace AbstractConfigGacela

Chemaclass:
## 📚  Description

The current logic that it's used by `Gacela::bootstrap()` and `gacela.php` it's similar but different... The goal is to normalise these two places in order to facilitate its usage and experience.

## 🔖 Changes

- Deprecate `AbstractConfigGacela`, in favor of `SetupGacela`

#### Current gacela.php style

```php
return static fn () => new class() extends AbstractConfigGacela {
    public function config(ConfigBuilder $configBuilder): void
    {
        $configBuilder->add('config/*.php', 'config/local.php');
    }
};
```

#### New style
```php
return static fn () => (new SetupGacela())
    ->setConfig(
        static function (ConfigBuilder $configBuilder): void {
            $configBuilder->add('config/*.php', 'config/local.php');
        }
    );

```

### DIFF
--- a/src/Framework/AbstractConfigGacela.php
+++ b/src/Framework/AbstractConfigGacela.php
@@ -4,29 +4,11 @@ declare(strict_types=1);
 
 namespace Gacela\Framework;
 
-use Gacela\Framework\Config\GacelaConfigBuilder\ConfigBuilder;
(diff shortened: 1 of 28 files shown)


Tito-Kati (approved):
♥

JesusValera (approved):

PR #128, opened 2022-03-26
Update php cs rules

Chemaclass:
## 🔖 Changes

- Composer update phpstan to latest 1.5
- Use PHPBench revs & iterations from phpbench.json config file
- Disable yoda style


JesusValera (approved):

Tito-Kati:
As a good Sith, I like Yoda well away from my code, good change!

PR #134, opened 2022-04-02
Improve mapping interfaces

Chemaclass:
## 🔖 Changes

Just internal refactoring.

- Instead of passing the `GacelaConfigFile` object, pass just the `mappingInterfaces` array to the `DependencyResolver`.
- Create `DependencyException` to group all exceptions using named-constructors.


src/Framework/ClassResolver/DependencyResolver/DependencyException.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public static function noParameterTypeFor(string $parameter): self
+    {
+        return new self("No parameter type for '{$parameter}'");
+    }
+
+    public static function unableToResolve(string $parameter, string $className): self
+    {
+        return new self("Unable to resolve [{$parameter}] from {$className}");
+    }
```
JesusValera, line 19: What about splitting this class into two?

1. A new class that extends from `InvalidArgumentException` class for `noParameterTypeFor()` and `unableToResolve()` methods
2. Another that extends from `RuntimeException` for the `mapNotFoundForClassName()` method

JesusValera (approved):
Nice 🎉

src/Framework/ClassResolver/DependencyResolver/DependencyException.php
Chemaclass: Nice idea. Feel free to do it yourself in this branch.

PR #142, opened 2022-04-23
DocBlockResolverAware feature

Chemaclass:
## 📚 Background

We implemented something similar "CustomServices" a few months ago (https://github.com/gacela-project/gacela/pull/84) but we removed it a few days later (https://github.com/gacela-project/gacela/pull/97) after realising the high complexity and lowering performance under it.

In this PR I try to implement again that concept, but keeping in mind offering a flexible possibility to define custom (and optionals) services using the docBlock from the class.

## 🔖 Changes

- These "DocBlockServices" have the Gacela autoload (so you can pass the Facade, Factory, or anything)
- Ideal for your infrastructure classes (for example, controllers or commands)
  - you need to `use DocBlockResolverAwareTrait`
  - and define in the class **docBlock the fully-qualified or relative class** that points to a particular method

## 🦤 Example

This `DocBlockResolverAwareTrait` is the new feature that allows you to define custom services without the need of inheritance or implementing any interface. The main benefits of this is that Gacela will create, cache and resolve its dependencies automatically for you.

```php
/**
 * @method CustomHelloRepository getRepository()
 */
final class HelloCommand extends Command
{
    use DocBlockResolverAwareTrait;

    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        $output->write($this->getRepository()->findNameName());

        return self::SUCCESS;
    }
}
```

##  🤔  Follow up idea

This feature could deprecate the `FacadeResolverAwareTrait` feature, because we wont need to implement `facadeClass()`. 

### Comparison

Instead of using the `FacadeResolverAwareTrait`:
```php
/**
 * @method Facade getFacade()
 */
final class TestHiCommand extends Command
{
    use FacadeResolverAwareTrait;

    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        $output->write($this->getFacade()->sayHi());

        return self::SUCCESS;
    }

    /** THIS method wont be necessary using `DocBlockResolverAwareTrait */
    protected function facadeClass(): string
    {
        return Facade::class;
    }
}
```

We could use the new `DocblockResolverAwareTrait`: 
```php
/**
 * @method Facade getFacade()
 */
final class TestHiCommand extends Command
{
    use DocBlockResolverAwareTrait;

    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        $output->write($this->getFacade()->sayHi());

        return self::SUCCESS;
    }
}
```

## 💡  Follow up
Maybe consider adding some file cache like we did here https://github.com/gacela-project/gacela/pull/120 if we see any performance problem because of this file IO to read the `use` statements.


JesusValera:
We are missing in `UseBlockParser` the scenario when the class is in the same namespace as the looking one?

src/Framework/ClassResolver/DocBlockService/UseBlockParser.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $needle = "{$className};";
+        $lines = array_filter(
+            explode(PHP_EOL, $phpCode),
+            static fn (string $l) => str_contains($l, $needle)
+        );
+        /** @psalm-suppress RedundantCast */
+        $lineSplit = (array)explode(' ', (string)reset($lines));
+
+        return rtrim($lineSplit[1] ?? '', ';');
```
JesusValera, line 35: @Chemaclass maybe we should check the line starts with `use ...`?

image

JesusValera (approved):

PR #146, opened 2022-05-05
gacela.php using callable with GacelaConfig as arg

Chemaclass:
## 📚 Description

Improve the UX when using gacela.php config file. Instead of returning an object, and use a chain of setters, use object methods like this:

```php
// gacela.php
return static function (GacelaConfig $config): void {
    $config
        // adding app configuration
        ->addAppConfig('config/*.php', 'config/local.php')

        // adding suffix types
        ->addSuffixTypeFacade('FacadeFromBootstrap')
        ->addSuffixTypeFactory('FactoryFromBootstrap')
        ->addSuffixTypeConfig('ConfigFromBootstrap')
        ->addSuffixTypeDependencyProvider('DependencyProviderFromBootstrap')

        // Mapping interfaces
        ->addMappingInterface(GreeterGeneratorInterface::class, IncorrectCompanyGenerator::class)
        ->addMappingInterface(CustomInterface::class, $config->getExternalService('CustomClassKey'));
};
```

> Inspired by the idea of simplifying the config setup https://getrector.org/blog/new-in-rector-012-introducing-rector-config-with-autocomplete Thanks to @TomasVotruba

## 🔖 Changes

- Allow using `GacelaConfig` as argument from a callable from gacela.php.
- This is backwards compatible with the current `GacelaSetup` fluent-setters design.

## 💡 Follow up ideas

Maybe this could deprecate the current `GacelaSetup` fluent-setters design in favor of the new `GacelaConfig` design.


TomasVotruba:
Cool to see it applied in practice 😊
Thanks for sharing

Chemaclass:
Those methods `addAppConfig()`, `addSuffixTypeFacade()`, `mapInterface()`, ... are the first things that came to my mind. Feel free to suggest better names, please @JesusValera @jenshaase @Tito-Kati

jenshaase:
PHP CS Fixer does something similar: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/config.rst

One drawback I think of is that the `GacelaConfig` class is not immutable. Wouldn't it be better to have a function that returns a new instance of `GacelaConfig`. So instead of 

```php
return static function (GacelaConfig $config): void {
// ...
}
```
Do this:
```php
return static function (GacelaConfig $config): GacelaConfig {
  return $config
}
```

However, for Phel this is not very interesting since we use the `GacelaSetup` class.

JesusValera:
> PHP CS Fixer does something similar: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/config.rst
> 
> One drawback I think of is that the `GacelaConfig` class is not immutable. Wouldn't it be better to have a function that returns a new instance of `GacelaConfig`. So instead of
> 
> ```
> return static function (GacelaConfig $config): void {
> // ...
> }
> ```
> 
> Do this:
> 
> ```
> return static function (GacelaConfig $config): GacelaConfig {
>   return $config
> }
> ```
> 
> However, for Phel this is not very interesting since we use the `GacelaSetup` class.

Interesting point of view @jenshaase.
But I would say the purpose of the `GacelaConfig` class is exactly this one, to mutate the state for each alteration.
I think it's easier and better this way, else, we should copy all different data (mapping interfaces, app config and suffixes) for each new item the user adds to the configuration.
Additionally, the scope of the `GacelaConfig` instance is only inside this closure.

What do you think? I am open for new perspectives :)

JesusValera:
This is the actual interface of `GacelaConfig`

image
There are so many methods, and most of them are for our internal purposes, the customer should see as less internal things as possible, I would avoid showing as many as possible those methods, in order to make the usage easier and avoid possible misinterpretations of the class itself, what do you think @Chemaclass?

It comes to my mind two alternatives:
1. A single `build()` method [we can mark it also as `@internals`] that returns an array with all information, and in the Gacela internals, we took what we want, or
2. Use Reflection to mark the classes `ConfigBuilder`, `SuffixTypesBuilder` and `MappingInterfacesBuilder` as public even if they are private (I don't like this approach, also probably we should drop the `final` keyword). But this would be totally transparent without `build()` or anything else to the final user.

---

Mockup of the first version:

image

Chemaclass:
Good Point. I am also not so happy about all those public methods. I wouldn't use reflection for this, The first option is better. Although, I am not 100% happy with that build option either, but it looks better as the current implemented. Feel free to push a commit with your suggestion directly to this branch, @JesusValera

Chemaclass:
@jenshaase regarding your suggestion embracing immutable for this `$config` object:
> ```php
> return static function (GacelaConfig $config): GacelaConfig {
>   return $config
> }
> ```
I think there is no real win having this approach, because the object it's being passed as reference as it will be pass as new object from Gacela point-of-view and it will be filled by the client. You can use both approaches; either keep return `void` or returning the modified `GacelaConfig`, making even possible an arrow function for this approach like this:
```php
return static fn (GacelaConfig $config) => $config
    ->addAppConfig('config/from-gacela-file.php')
    ->addMappingInterface(CustomInterface::class, $config->getExternalService('CustomClassFromExternalService'))
    ->addSuffixTypeFacade('FacadeFromGacelaFile')
    ->addSuffixTypeFactory('FactoryFromGacelaFile')
    ->addSuffixTypeConfig('ConfigFromGacelaFile')
    ->addSuffixTypeDependencyProvider('DependencyProviderFromGacelaFile');
```

---

> However, for Phel this is not very interesting since we use the GacelaSetup class.

Well, I see this new concept as a new approach that could deprecated the current SetupGacela (in order to simplify the setup itself...). Here you can see a comparison from the current vs the new design:

```php
# phel-lang/phel:21

// CURRENT
$setupGacela = (new SetupGacela())
    ->setConfig(static function (ConfigBuilder $configBuilder): void {
        $configBuilder->add('phel-config.php', 'phel-config-local.php');
    });
// NEW
$setupGacela = static fn (GacelaConfig $config) => $config
    ->addAppConfig('phel-config.php', 'phel-config-local.php');

Gacela::bootstrap($projectRootDir, $setupGacela);
```

src/Framework/Bootstrap/AbstractSetupGacela.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @deprecated in favor of `externalServices()`
+     *
+     * @return array<string,mixed>
+     */
+    public function globalServices(): array
+    {
+        return $this->externalServices();
+    }
```
JesusValera, line 55: This method was deprecated one month ago on version [0.16.0](https://github.com/gacela-project/gacela/blob/0.16.0/src/Framework/Setup/SetupGacela.php#L92).
Maybe we could drop it in this release @Chemaclass?

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigFromBootstrapFactory.php
```diff
@@ -35,25 +35,16 @@ public function createGacelaFileConfig(): GacelaConfigFileInterface
 
     private function createConfigBuilder(): ConfigBuilder
     {
-        $configBuilder = new ConfigBuilder();
-        $this->setup->config($configBuilder);
-
-        return $configBuilder;
+        return $this->setup->buildConfig(new ConfigBuilder());
```
JesusValera, line 38: Nice 👍🏼

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigUsingGacelaPhpFileFactory.php
```diff
@@ -37,13 +40,16 @@ public function createGacelaFileConfig(): GacelaConfigFileInterface
...
         /** @var SetupGacelaInterface|callable $setupGacelaFile */
         $setupGacelaFile = $this->fileIo->include($this->gacelaPhpPath);
+
         if (is_callable($setupGacelaFile)) {
-            trigger_deprecation('Gacela', '0.15', 'Return a SetupGacelaInterface instance directly. The callable option will be removed in the next version.');
+            $gacelaConfig = new GacelaConfig($this->setup->externalServices());
+            $setupGacelaFile($gacelaConfig);
+            $setupGacela = SetupGacela::fromGacelaConfig($gacelaConfig);
+        } else {
+            $setupGacela = $setupGacelaFile;
```
JesusValera, line 49: Will we drop in the future the `(SetupGacela())->setXXX()` construction?

src/Framework/Setup/AbstractSetupGacela.php
```diff
@@ -4,50 +4,11 @@
...
 namespace Gacela\Framework\Setup;
 
-use Gacela\Framework\Config\GacelaConfigBuilder\ConfigBuilder;
-use Gacela\Framework\Config\GacelaConfigBuilder\MappingInterfacesBuilder;
-use Gacela\Framework\Config\GacelaConfigBuilder\SuffixTypesBuilder;
+use Gacela\Framework\Bootstrap\AbstractSetupGacela as BootstrapAbstractSetupGacela;
 
-abstract class AbstractSetupGacela implements SetupGacelaInterface
+/**
+ * @deprecated in favor of \Gacela\Framework\Bootstrap\AbstractSetupGacela
```
JesusValera, line 10: What about defining the version we will remove the deprecation class/method in the PhpDoc itself?

src/Framework/Bootstrap/AbstractSetupGacela.php
Chemaclass: Good idea. Feel free to do this in a follow up PR.

src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigUsingGacelaPhpFileFactory.php
Chemaclass: Good idea, but it will require some minor changes. Please, apply these changes in a follow PR.

src/Framework/Setup/AbstractSetupGacela.php
Chemaclass: I would prefer don't. The "deprecated" tag already tell you that you shouldn't use it and it will be removed in a future version. So it's up to the dev to avoid using the deprecated and use the new ones.

JesusValera (approved):

PR #153, opened 2022-05-19
Removed bin/gacela util from this repo

Chemaclass:
## 📚 Description

The `symfony/console` has some dependencies that gacela doesn't really need:
```json
    "require": {
        "php": ">=7.2.5",
        "symfony/deprecation-contracts": "^2.1|^3",
        "symfony/polyfill-mbstring": "~1.0",
        "symfony/polyfill-php73": "^1.9",
        "symfony/polyfill-php80": "^1.16",
        "symfony/service-contracts": "^1.1|^2|^3",
        "symfony/string": "^5.1|^6.0"
    },
```

That dependency was introduced ONLY to be able to run "`vendor/bin/gacela` code generator commands" which aren't really used always in all projects using gacela.

## 🔖 Changes

- Remove the `symfony/console` dependency
- Move away the CodeGenerator module to its own repository: https://github.com/gacela-project/gacela-cli

This way the real dependencies for gacela will be just these:
```json
    "require": {
        "php": ">=7.4",
        "ext-json": "*"
    },
```


JesusValera (approved):
Interesting approach, one question:
Does it make sense to have the `Framework` folder now?
image

Chemaclass:
Yes, because even if we split the repos, they still share the namespace Gacela. So, I like to keep the Framework level.

PR #155, opened 2022-05-21
Allow gacela anon-classes using parent methods

Chemaclass:
## 📚 Description

Currently, if you use `AnonymousGlobal::overrideExistingResolvedClass()` to override a real `Factory` because you are interested on overriding some method(s) in your child class (potentially anonymous class), for example, the method `getConfig()` from the parent class wont be called. Instead you'll face an Exception.

The reason is that currently, when using `getConfig()` it tries to resolve it from the current caller. And in the case of the anon-class, the current caller is the file/class name where the anon-class was created.

The solution is to check the parent class namespace in case there was no resolver class found for a particular caller. 

## 🔖 Changes

- Try recursively once more with its parent class on `doResolve()` if the caller was an object.
- Moved `Gacela/Framework/Config` to `Gacela/Framework/Config/Config`
  - Why? Because `test_missing_config_module` was failing due to the fact that `Config` is a valid name next to the `AbstractFactory`. So we have to put it in another directly. Anyway, it fits better in that place. Also, it's an internal class, so no broken changes for the public API.


src/Framework/ClassResolver/AbstractClassResolver.php
```diff
@@ -44,6 +45,14 @@ public function doResolve($caller): ?object
 
         $resolvedClassName = $this->findClassName($classInfo);
         if ($resolvedClassName === null) {
+            // Try once more with its parent class
```
JesusValera, line 48: I think we should not try just once but recursively till there are no more parents.

Chemaclass: That's exactly what it does... maybe the comment it's misleading. I will improve it.

JesusValera (approved):

PR #157, opened 2022-05-23
Add AbstractClassResolver::resetCache()

JesusValera:
## 📚 Description

Create `AbstractClassResolver::resetCache()` method in order to remove the cache, this function is focus for tests when using `@dataproviders` for example.

### DIFF
--- a/src/Framework/ClassResolver/AbstractClassResolver.php
+++ b/src/Framework/ClassResolver/AbstractClassResolver.php
@@ -24,6 +24,9 @@ abstract class AbstractClassResolver
 
     private ?InstanceCreator $instanceCreator = null;
 
+    /**
+     * @internal just for testing purposes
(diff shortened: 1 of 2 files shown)


src/Framework/ClassResolver/AbstractClassResolver.php
```diff
@@ -24,6 +24,11 @@ abstract class AbstractClassResolver
 
     private ?InstanceCreator $instanceCreator = null;
 
+    public static function resetCache(): void
```
Chemaclass, line 27: What do you think about adding the flag `@internal just for testing purposes` (or something similar) to make the function not API friendly, as a warning for other people who might be interested to use it? I don't know. What do you think, @JesusValera?

Chemaclass (approved):

PR #164, opened 2022-06-02
Doc block resolver aware with no parent

Chemaclass:
## 📚 Description

When `DocBlockResolverAwareTrait` uses `parent::class` without checking if there is or not a parent class then you might face this  ```Error: Cannot use "parent" when current class scope has no parent```.

## 🔖 Changes

- Check if the current class has a parent before accessing the `parent::class` .

### DIFF
--- a/src/Framework/DocBlockResolverAwareTrait.php
+++ b/src/Framework/DocBlockResolverAwareTrait.php
@@ -28,22 +28,25 @@ trait DocBlockResolverAwareTrait
      */
     public function __call($method, $parameters = [])
     {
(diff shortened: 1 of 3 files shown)


JesusValera (approved):
Nice catch!
How did you find this problem? (it worked for me 🤔)
If this bug is critical, maybe should we create a `0.21.1` version, what do you think?

Chemaclass:
No, it's not really critical. Actually, there is another tricky thing here that I wasn't sure how to solve or if it make sense... Let's pair on this topic this weekend :)

PR #167, opened 2022-06-07
Rename $config->resetCache() to $config->setCacheEnabled()

JesusValera:
## 📚 Description

Rename the GacelaConfig method from `$config->resetCache(bool)` to `$config->setCacheEnabled(bool)`.
The default value is now `true`.

## 🪱 Usage example

image

### DIFF
--- a/src/Framework/Gacela.php
+++ b/src/Framework/Gacela.php
@@ -24,7 +24,7 @@ final class Gacela
             ? SetupGacela::fromCallable($configFn)
             : new SetupGacela();
 
-        if ($setup->isCacheEnabled()) {
+        if (!$setup->isCacheEnabled()) {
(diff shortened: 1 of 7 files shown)


src/Framework/Gacela.php
```diff
@@ -24,7 +24,7 @@ public static function bootstrap(string $appRootDir, Closure $configFn = null):
             ? SetupGacela::fromCallable($configFn)
             : new SetupGacela();
 
-        if ($setup->isResetCache()) {
+        if ($setup->isCacheEnabled()) {
```
Chemaclass, line 27:
```suggestion
        if (!$setup->isCacheEnabled()) {
```

Chemaclass (approved):

PR #169, opened 2022-06-10
Add PhpDoc to $parentReturn var

JesusValera:
## 📚 Description

The old tag `@psalm-suppress ParentNotFound` was not necessary anymore because in the previous `if`, we check if the current class has already some parent class.

In another way, from time to time the Psalm complain because the `$parentReturn` is not a `string-class` but a simple string, with this PhpDoc Psalm is not complaining anymore.

### DIFF
--- a/src/Framework/DocBlockResolverAwareTrait.php
+++ b/src/Framework/DocBlockResolverAwareTrait.php
@@ -54,7 +54,7 @@ trait DocBlockResolverAwareTrait
         }
 
         if ($this->hasParentClass()) {
-            /** @psalm-suppress ParentNotFound */
+            /** @var class-string $parentReturn */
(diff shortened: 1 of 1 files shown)


src/Framework/DocBlockResolverAwareTrait.php
```diff
@@ -54,7 +54,7 @@ public function __call($method, $parameters = [])
         }
 
         if ($this->hasParentClass()) {
-            /** @psalm-suppress ParentNotFound */
+            /** @var class-string $parentReturn */
             $parentReturn = parent::__call($method, $parameters);
```
Chemaclass, line 58: The parentReturn is not a class-string, but a ?mixed. I am not sure why you did this change 🤔

Chemaclass (changes requested):

PR #171, opened 2022-06-24
Group gacela cache files inside a cache directory

Chemaclass:
## 📚 Description

Currently the gacela cache files are created under the root project dir.

## 🔖 Changes

- Create a directory "cache" on the fly in case it does not exists.
- Store inside the "cache" directory the cache files.
- Allow enable/disable the cache from the config files directly.


JesusValera (approved):
I'm fine with this default path, but I would even consider the possibility to define the path in the ConfigBuilder, what do you think?

PR #175, opened 2022-06-24
Fix setCacheDirectory() with nested dir levels

Chemaclass:
## 📚 Description

Currently, `$config->setCacheDirectory('custom-caching-dir');` doesn't work if you points to a non existing directory with multiple nested level; for example `'custom/caching-dir'`.

## 🔖 Changes

- Fix nested directly levels for the cache directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 0.23.1
+#### 2022-06-25
+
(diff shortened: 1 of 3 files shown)


JesusValera (approved):

JesusValera:
Nice 🙌🏼 

Please, don't forget to update the `CHANGELOG` file.

PR #181, opened 2022-07-07
Load both gacela.php and gacela-xxx.php files

JesusValera:
## 📚 Description

Having a `gacela.php` and a `gacela-prod.php` file.

If we define the `APP_ENV` as `dev`
1. `gacela.php` should be loaded

When we define the `APP_ENV` as `prod`
1. `gacela.php` should be loaded
2. `gacela-prod.php` should be loaded right after (overriding the keys if repeated)


Chemaclass (approved):
Great idea! 💡

PR #182, opened 2022-07-09
Increase Code Coverage excluding some files

JesusValera:
## 📚 Description

Increase Code Coverage from 87% to 94% excluding some files from `src/CodeCoverage` module.

We already have [excellent `feature` tests for the module](https://github.com/gacela-project/gacela/tree/master/tests/Feature/CodeGenerator), so, those classes are being "tested" indirectly through PHP `exec()` command. Still, unfortunately, they are not included in the PHPUnit coverage, and because of that, the code coverage drops 7 points.

I think the best alternative is just ignore them 😄

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -34,5 +34,9 @@
         <include>
             <directory suffix=".php">src</directory>
         </include>
+        <exclude>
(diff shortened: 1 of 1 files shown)


Chemaclass (approved):
Good idea. Thanks

Issue #178, opened 2022-06-30
Allow adding config key-values from GacelaConfig


Chemaclass:
## Current behaviour

```php
Gacela::bootstrap(__DIR__, static function (GacelaConfig $config): void {
    $config->addAppConfig('config/default.php');
});
```

## Desired behaviour

```php
Gacela::bootstrap(__DIR__, static function (GacelaConfig $config): void {
    # Currently, we can define a path from where to load config files
    $config->addAppConfig('config/*.php');

    # We want to allow also defining key-values directly from the bootstrap as well.
    # This is pretty handy for a quick bootstrapping, trying values, or even for testing.
    # These values should override the values defined in the appConfig files. 
    $config->addAppConfigKeyValues([ // this is an idea
        'key-1' => 'value-1',
        'key-2' => 'value-2',
    ]);
});
```

Anas-debug:
I would like to take this issue.

Chemaclass:
Sure! Go for it. Let me know if you need any help.

Chemaclass:
After talking to @Anas-debug, we agreed this might be to complex as a "good first issue", so I will take it over. I provided him another simpler issue to start contributing.

PR #183, opened 2022-07-13
Allow adding config key-values from GacelaConfig

Chemaclass:
## 📚 Description

Issue: https://github.com/gacela-project/gacela/issues/178

## 🔖 Changes

- Add method: `GacelaConfig->addAppConfigKeyValue()`

### DIFF
--- a/src/CodeGenerator/Domain/CommandArguments/CommandArgumentsParser.php
+++ b/src/CodeGenerator/Domain/CommandArguments/CommandArgumentsParser.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Gacela\CodeGenerator\Domain\CommandArguments;
 
 use InvalidArgumentException;
+
 use function count;
 
 final class CommandArgumentsParser implements CommandArgumentsParserInterface
--- a/src/CodeGenerator/Domain/FilenameSanitizer/FilenameSanitizer.php
+++ b/src/CodeGenerator/Domain/FilenameSanitizer/FilenameSanitizer.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Gacela\CodeGenerator\Domain\FilenameSanitizer;
 
 use RuntimeException;
+
 use function count;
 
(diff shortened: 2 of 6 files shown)


tests/Feature/Framework/ResolveDifferentProjectNamespaces/FeatureTest.php
```diff
@@ -22,7 +23,7 @@ final class FeatureTest extends TestCase
     public function setUp(): void
     {
         Gacela::bootstrap(__DIR__, static function (GacelaConfig $config): void {
-            $config->addAppConfig('config/default.php');
+            $config->addAppConfigKeyValue(GacelaCache::KEY_ENABLED, false);
```
JesusValera, line 26: Not sure if I understand the PR.

```php
# default.php
return [
    GacelaCache::KEY_ENABLED => false,
];

// is the same than?

# index.php
    $config->addAppConfigKeyValue(GacelaCache::KEY_ENABLED, false);
}
```
If so, in the `addAppConfigKeyValue()` method, the `GacelaCache::KEY_ENABLED` is the key and the `false` is the value?

Then, how could I add multiple keys? Like this? 🤔 
```php
$config->addAppConfigKeyValue(GacelaCache::KEY_ENABLED, false);
$config->addAppConfigKeyValue('some key', 'some value');
$config->addAppConfigKeyValue('other key', 'other value');
```

What about this proposal @Chemaclass? I would find even more interested to do something like

```php
$config->addAppConfigKeyValue([
    GacelaCache::KEY_ENABLED => false,
    'some key' => 'some value',
    'other key' => 'other value',
]);
```

Chemaclass: Yes, that would be another way. Do you want to do it in a follow up PR, @JesusValera?

JesusValera: Sure thing :)

JesusValera (approved):

PR #189, opened 2022-07-31
Remove dead code from DocBlockResolverAwareTrait

JesusValera:
## 📚 Description

I remember we faced some errors when in Laravel a `Controller` class, a userland use the `DocBlockResolverAwareTrait`.
This error was happening because both (the trait and the controller) implemented the magic `__call()` method. In some scenarios, when the trait was implemented, it was overriding the `__call()` from the controller facing some undesirable effects.

It seems that this error is not happening anymore in Laravel (I could not reproduce this scenario), so, I decided to remove this _dead_ code from now.


Chemaclass (approved):
Less is more! Nice catch 👌

PR #196, opened 2022-09-25
Use easy-coding-standard instead of php-cs-fixer

Chemaclass:
## 📚 Description

I recently discover this `easy-coding-standard` (by @TomasVotruba), and I think it's much simpler and straight forward than maintaining our custom `.php-cs-fixer.dist.php`.

## 🔖 Changes

- Remove `php-cs-fixer` in favor of `easy-coding-standard`


JesusValera (changes requested):
I tried to learn the `ecs`, but I found it more complex than `php-cs-fixer`.
If you are able to add all rules from the `php-cs-fixer` to `ecs`, that would be nice but in your PR you are not adding any rule at all 😄

Please, add them or close the current PR, right now it is incomplete, e.g:
If I move the private function before of another public function, and add the `\` to all native functions and run the `ecs fixer` command, it says everything is ok.

image

PR #197, opened 2022-09-27
Use InMemoryCache and make the FileCache as a profiler feature

Chemaclass:
## 📚 Description

I discovered that the gacela file cache didn't work as expected, and that it didn't really help at all improving the performance. However, the feature of generating the internal mapping between the gacela resolvable types and the final class that it's used can be interesting as a new profiling feature.

## 🔖 Changes

- Add (optional) gacela file profiler. By default disabled.
- Remove gacela file cache. Instead, use in memory cache always.
- Remove `gacela cache:clear` command, because there is no file cache anymore.


JesusValera (approved):
🎉 Don't forget to update the web with this new `profiler` stuff 😄

PR #199, opened 2022-09-30
Fix PHPStan error should return class-string but returns string

JesusValera:
## 📚 Description

Add to `phpstan.neon` the rule to ignore when a `class-string` should be returned but returns a `string` and clean a little the code 😄


Chemaclass (approved):
Good idea! 👏

PR #201, opened 2022-10-01
Add full compatibility for windows support

Chemaclass:
## 📚 Description

I just realised that there are some spots where gacela doesn't work for windows. There is a failing test, but I wonder if that's all 🤔 What about the hardcoded `/` instead of using the `DIRECTORY_SEPARATOR` PHP constant? We need to check that.

## 🔖 Changes

- Add operating-systems (ubuntu,windows,macos) to the CI matrix
- WIP: Fix the broken tests


Chemaclass:
Thanks, @JesusValera, for your changes so far! 🥇

.github/workflows/tests.yml
```diff
@@ -20,6 +20,7 @@ jobs:
         php-version: [ "8.0" , "8.1", "8.2" ]
         operating-system:
           - "ubuntu-latest"
+#          - "windows-latest" # TODO: Windows is not working in the CI at the moment for some reason
```
JesusValera, line 23: I downloaded and ran the tests using a Windows machine, all tests are green now.

phpunit-gacela-windows-error

But the `GitHub Actions` is failing, in any case, I think this commit brings value to Windows users, and we can investigate in the future why the CI for Windows is not working. Also, I am unable to reproduce this error 🤔 

image

Trace: https://github.com/gacela-project/gacela/actions/runs/5171780050/jobs/9315512655

Chemaclass: Yes, let's merge it with your changes -as they provide value as it is-, and we can continue the investigation about why is the Windows workflow not working properly as they are in your windows machine.

PR #203, opened 2022-10-04
Allow using Interface suffix for DocBlockResolver

Chemaclass:
## 📚 Description

Allow using an interface of your Facade (or any Service) specified as DocBlock resolvable, for example:

```php
/**
 * @method FakeServiceInterface getService()
 */
final class DummyDocBlockResolverAware
{
    use DocBlockResolverAwareTrait;
}
```

## 🔖 Changes

- On `getClassFromDoc()` add a last condition checking if the class name has `Interface` on it, and try finding a class name without the interface suffix.

### Things to consider

This feature assumes that:
- ...your interface will be on the same namespace as the concrete implementation
- ...the concrete implementation has the same name as the interface

> I would personally use the concrete class name in the DocBlock resolvable, but it might be useful for other scenarios.

### DIFF
--- a/src/Framework/DocBlockResolver/DocBlockResolver.php
+++ b/src/Framework/DocBlockResolver/DocBlockResolver.php
@@ -21,6 +21,8 @@ use function is_string;
 
 final class DocBlockResolver
 {
+    private const INTERFACE_SUFFIX = 'Interface';
+
     /** @var array<string,string> [fileName => fileContent] */
     private static array $fileContentCache = [];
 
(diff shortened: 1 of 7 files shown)


JesusValera (changes requested):
Interesting PR @Chemaclass but I don't like the idea.

Implementing this PR will bring a lot of unnecessary magic to the project.
As less magic as possible, the better in my opinion.

Furthermore I find some negative cases if we implement something like this (to depend on the interface name to see the referring class without the `Interface` word on it).

Some ideas that pop up in my mind:
1. Not everybody uses the `Interface` suffix when creating interface classes
2. Maybe people use this system to name the interfaces with the `Interface` suffix, but the classes that implement this interface don't follow this schema, eg:
  - `DatabaseInterface.php`
  - `MySQLDatabase.php`
  - `PostgreSQLDatabase.php`
3. Having the previous example where there is a match between the name of the interface and the class that implements this interface, but, what about?
  - Depending on the environment we really want to load a different class?
  - We want to load another class (or maybe an Adapter class) for any other reason?
In the latter example, we are forced to use the original class (the framework will auto-resolve the incorrect class and we cannot do anything).
4. Probably we have an interface and a class with the same name as the interface, BUT this class is in another namespace, even in this situation the framework won't be able to resolve the class because won't find it
5. In case we want to rename/move the classes, the project will fail...

There are multiple problems if you only trust the naming.

My recommendation would be to be able to define in the `DocBlockResolver` an interface class, but Gacela would auto-resolve the original class from the `mappingInterfaces` in the `SetupGacela` file.
I am not sure if this is implemented, any idea @Chemaclass? What do you think about my message? 😄

PR #212, opened 2022-10-23
Remove the profiler feature

Chemaclass:
## 🔖 Changes

- Remove the `profiler feature` because the `file cache` has the same behaviour

With this we remove a lot of unnecessary complexity from the project 🥳


JesusValera (approved):
+100

PR #214, opened 2022-10-27
Fix: Do not use deprecated set-output command

localheinz:
This pull request

- [x] stops using the deprecated `set-output` command

💁‍♂️ For reference, see https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.


Chemaclass (approved):
I didn't know this. Thanks!

localheinz:
Thank you, @Chemaclass and @JesusValera!

PR #215, opened 2022-10-27
Adding an event listeners system

Chemaclass:
## 📚 Description

Adding event listeners mechanism to be able to react on events when gacela resolves class names, so the user can do whatever they want: for example debugging, logging, tracing…

## 🔖 Changes

- Added `public function registerListener(string $event, callable $listener): void`
- Trigger events when gacela resolves class names:
  - ResolvedClassCachedEvent
  - ResolvedClassCreatedEvent
  - ResolvedClassTriedFromParentEvent
  - ResolvedCreatedDefaultClassEvent

### 🧪 Example 

```php
<?php declare(strict_types=1);

namespace GacelaData\Events;

require __DIR__ . '/../../vendor/autoload.php';

# This is an example; you can check the feature tests for a deeper look
use Gacela\Framework\Bootstrap\GacelaConfig;
use Gacela\Framework\EventListener\ClassResolver\ResolvedClassCachedEvent;
use Gacela\Framework\EventListener\ClassResolver\ResolvedClassCreatedEvent;
use Gacela\Framework\EventListener\ClassResolver\ResolvedClassTryFormParentEvent;
use Gacela\Framework\EventListener\ClassResolver\ResolvedDefaultClassEvent;
use Gacela\Framework\EventListener\GacelaEventInterface;
use Gacela\Framework\Gacela;

$renderEvent = static function (GacelaEventInterface $event): void {
    echo sprintf(
        "%s - %s\n",
        \get_class($event),
        $event->classInfo()->getResolvableType()
    );
};

Gacela::bootstrap(__DIR__, static function (GacelaConfig $config) use ($renderEvent): void {
    $config->registerListener(ResolvedClassCachedEvent::class, $renderEvent);
    $config->registerListener(ResolvedClassCreatedEvent::class, $renderEvent);
    $config->registerListener(ResolvedClassTryFormParentEvent::class, $renderEvent);
    $config->registerListener(ResolvedDefaultClassEvent::class, $renderEvent);
});

# the facade uses the factory, which needs to be created while resolving it
$facade = new Mod\Facade();
$facade->doString(); // ResolvedClassCreatedEvent - Factory

# this time, the factory was cached in the previous call
$facade = new Mod\Facade();
$facade->doString(); // ResolvedClassCachedEvent - Factory

# if there is no Config file, Gacela will create one default-empty for that particular module
$factory = new Mod\Factory();
$factory->getConfig();
// ResolvedClassTryFormParentEvent - Config
// ResolvedDefaultClassEvent - Config

# even the default-empty Config will be cached this time as well
$factory = new Mod\Factory();
$factory->getConfig();
// ResolvedClassTryFormParentEvent - Config
// ResolvedClassCachedEvent - Config
```

### DIFF
--- a/src/Framework/Bootstrap/GacelaConfig.php
+++ b/src/Framework/Bootstrap/GacelaConfig.php
@@ -33,6 +33,9 @@ final class GacelaConfig
     /** @var array<string,mixed> */
     private array $configKeyValues = [];
 
+    /** @var array<string,list<callable>> */
+    private array $listeners = [];
+
     /**
      * @param array<string,class-string|object|callable> $externalServices
      */
@@ -211,6 +214,13 @@ final class GacelaConfig
         return $this;
     }
 
+    public function addEventListener(string $listener, callable $callable): self
+    {
+        $this->listeners[$listener][] = $callable;
+
+        return $this;
+    }
+
     /**
      * @internal
      *
@@ -224,6 +234,7 @@ final class GacelaConfig
      *     file-cache-directory: string,
      *     project-namespaces: list<string>,
      *     config-key-values: array<string,mixed>,
+     *     listeners: array<string,list<callable>>,
      * }
      */
     public function build(): array
@@ -238,6 +249,7 @@ final class GacelaConfig
             'file-cache-directory' => $this->fileCacheDirectory,
             'project-namespaces' => $this->projectNamespaces,
             'config-key-values' => $this->configKeyValues,
+            'listeners' => $this->listeners,
         ];
     }
 }
--- a/src/Framework/Bootstrap/GacelaConfig.php
+++ b/src/Framework/Bootstrap/GacelaConfig.php
@@ -35,8 +35,8 @@ final class GacelaConfig
 
     private bool $areEventListenersEnabled = true;
 
-    /** @var array<string,list<callable>> */
-    private array $eventListeners = [];
+    /** @var array<class-string,list<callable>> */
+    private array $listenersPerEvent = [];
 
     /**
      * @param array<string,class-string|object|callable> $externalServices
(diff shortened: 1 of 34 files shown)


src/Framework/Bootstrap/GacelaConfig.php
```diff
@@ -214,6 +216,13 @@ public function addAppConfigKeyValues(array $config): self
         return $this;
     }
 
+    public function disableEventListeners(): self
```
JesusValera, line 219: Could you please @Chemaclass add some PhpDoc to this method? :D

src/Framework/ClassResolver/AbstractClassResolver.php
```diff
@@ -156,4 +171,17 @@ private function createDefaultGacelaClass(): ?object
                 return null;
         }
     }
+
+    private function triggerEvent(GacelaEventInterface $event): void
```
JesusValera, line 175: I think it makes sense to move this logic to a new class with a `handle/trigger` method/whatsoever.
This way, we are not coupled to the `doResolve()` and we can add any other listener in the application, what do you think?

tests/Feature/Framework/ListeningEvents/DisableListenersTest.php
```diff
@@ -0,0 +1,41 @@
...
+        $inMemoryEvents = [];
+
+        Gacela::bootstrap(__DIR__, static function (GacelaConfig $config) use (&$inMemoryEvents): void {
+            $config->disableEventListeners();
+
+            $config->addEventListener(
+                GacelaClassResolverListener::class,
+                static function (GacelaEventInterface $event) use (&$inMemoryEvents): void {
+                    $inMemoryEvents[] = $event;
+                    throw new RuntimeException('This code should never be called');
```
JesusValera, line 28: Not sure if really makes sense this exception here, I suggest replacing it by a comment in the line before 🤔

JesusValera (commented):
Some suggestions 😃

JesusValera (approved):

PR #216, opened 2022-10-30
Add ConfigReaderEvents

JesusValera:
## 📚 Description

Dispatch the event `ReadPhpConfigEvent` when a PHP-config-file is being read.


src/Framework/Config/ConfigReader/PhpConfigReader.php
```diff
@@ -52,4 +57,9 @@ private function canRead(string $absolutePath): bool
 
         return $extension === 'php' && file_exists($absolutePath);
     }
+
+    private function triggerEvent(GacelaEventInterface $event): void
```
Chemaclass, line 61: Can we rename this to `dispatchEvent()`?

Chemaclass (approved):

src/Framework/Config/ConfigReader/PhpConfigReader.php
JesusValera: Absolutely ! 👍 💯

PR #217, opened 2022-10-30
Add GacelaConfig::registerGenericListener()

Chemaclass:
## 📚 Description

We added the possibility to register a listener to a particular event https://github.com/gacela-project/gacela/pull/215 
It would be useful to be able to register a listener to all posible events.

## 🔖 Changes

- Added `registerGenericListener()` to be able to register a generic listener for all events.
```php
GacelaConfig::registerGenericListener(callable $listener): void
```

## 🧪 A visual example
```php
$renderEvent = static function (GacelaEventInterface $event): void {
    echo sprintf("%s - %s\n", \get_class($event), $event->toString());
};

Gacela::bootstrap(__DIR__, static function (GacelaConfig $config) use ($renderEvent): void {
    // All of these individual specific listeners
    $config->registerSpecificListener(ResolvedClassCachedEvent::class, $renderEvent);
    $config->registerSpecificListener(ResolvedClassCreatedEvent::class, $renderEvent);
    $config->registerSpecificListener(ResolvedClassTriedFromParentEvent::class, $renderEvent);
    $config->registerSpecificListener(ResolvedCreatedDefaultClassEvent::class, $renderEvent);
    $config->registerSpecificListener(ReadPhpConfigEvent::class, $renderEvent);

    // would be the same as registering this generic listener
    $config->registerGenericListener($renderEvent);
});
```

## 🙇🏼  Especial thanks 

Credits to @matthiasnoback. I got the inspiration for this implementation thanks to the "[Layers, ports, and adapters workshop](https://github.com/matthiasnoback/layers-ports-and-adapters-workshop)" I did with him last year.

### DIFF
--- a/src/Framework/Bootstrap/GacelaConfig.php
+++ b/src/Framework/Bootstrap/GacelaConfig.php
@@ -37,6 +37,9 @@ final class GacelaConfig
 
     private bool $areEventListenersEnabled = true;
 
+    /** @var list<callable> */
+    private array $genericListeners = [];
+
     /** @var array<class-string,list<callable>> */
     private array $listenersPerEvent = [];
 
@@ -229,12 +232,20 @@ final class GacelaConfig
     }
 
     /**
+     * Register a generic listener when any event happens.
+     */
+    public function registerGenericListener(callable $listener): void
+    {
+        $this->genericListeners[] = $listener;
+    }
+
+    /**
      * Register a listener when some event happens.
      *
      * @param class-string $event
      * @param callable(GacelaEventInterface):void $listener
      */
-    public function registerListener(string $event, callable $listener): void
+    public function registerSpecificListener(string $event, callable $listener): void
     {
         $this->listenersPerEvent[$event][] = $listener;
     }
@@ -253,6 +264,7 @@ final class GacelaConfig
      *     project-namespaces: list<string>,
      *     config-key-values: array<string,mixed>,
      *     are-event-listeners-enabled: bool,
+     *     generic-listeners: list<callable>,
      *     listeners-per-event: array<class-string,list<callable>>,
      * }
      */
@@ -269,6 +281,7 @@ final class GacelaConfig
             'project-namespaces' => $this->projectNamespaces,
             'config-key-values' => $this->configKeyValues,
             'are-event-listeners-enabled' => $this->areEventListenersEnabled,
+            'generic-listeners' => $this->genericListeners,
             'listeners-per-event' => $this->listenersPerEvent,
         ];
     }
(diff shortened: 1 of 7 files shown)


src/Framework/Bootstrap/GacelaConfig.php
```diff
@@ -228,13 +231,21 @@ public function disableEventListeners(): self
...
     /**
      * Register a listener when some event happens.
      *
      * @param class-string $event
      * @param callable(GacelaEventInterface):void $listener
      */
-    public function registerListener(string $event, callable $listener): void
+    public function registerSpecificListener(string $event, callable $listener): void
     {
         $this->listenersPerEvent[$event][] = $listener;
```
JesusValera, line 250: What about renaming `listenersPerEvent` to `specificListeners`?

src/Framework/Bootstrap/GacelaConfig.php
```diff
@@ -228,13 +231,21 @@ public function disableEventListeners(): self
         return $this;
     }
 
+    /**
+     * Register a generic listener when any event happens.
```
JesusValera, line 235: Maybe a better description would be helpful (take a look at the review comment), I guess more people will have the same questions as I do 😃

src/Framework/Bootstrap/GacelaConfig.php
```diff
@@ -253,6 +264,7 @@ public function registerListener(string $event, callable $listener): void
      *     project-namespaces: list<string>,
      *     config-key-values: array<string,mixed>,
      *     are-event-listeners-enabled: bool,
+     *     generic-listeners: list<callable>,
      *     listeners-per-event: array<class-string,list<callable>>,
```
JesusValera, line 268: Same here with the name.

JesusValera (commented):
Not sure if I understand properly.
Are you forced to use the `GacelaEventInterface` in order to capture any other class that implements `GacelaEventInterface`? Can you use a specific event (eg: `ResolvedClassCachedEvent` class) instead of the `GacelaEventInterface`? In this scenario, what happens?

Chemaclass:
> Not sure if I understand properly. Should you use the `GacelaEventInterface` in order to capture any other class that implements `GacelaEventInterface`?
> 
> Can you use a specific event (eg: `ResolvedClassCachedEvent` class) instead of the `GacelaEventInterface`? In this scenario, what happens?

Yes, you must use `GacelaEventInterface` on the event that you received in the generic listener, because that's the global minimum "contract level" of all gacela events. Something more generic would ended up in an error on runtime.

If you would be interested in a particular event within a generic listener you would have to use a casting to another concrete.

src/Framework/Bootstrap/GacelaConfig.php
Chemaclass: I added some more doc lines below.

JesusValera (approved):
I see, great!

PR #221, opened 2022-11-02
Refactor dispatchEvent into EventDispatchingCapabilities trait

Chemaclass:
## 📚 Description

In every place where we want to dispatch events we are repeating the same code.

## 🔖 Changes

- Create a trait `EventDispatchingCapabilities` which has the method `dispatchEvent()` so it will simplify the way how to dispatch events.


localheinz (approved):
Although I am wary of using traits (as they often make code more difficult to understand), this looks good to me.

JesusValera (approved):

JesusValera:
I think exactly the same as @localheinz 👍🏼

Chemaclass:
I think exactly as you two @localheinz and @JesusValera. I am not a fan of traits for that exact reason, but in this case, I think it's pretty handy and useful 💡 Thanks for your feedback!

PR #222, opened 2022-11-04
Add overrideEventDispatcher() to SetupGacela

Chemaclass:
## 📚 Description

With this you will be able to override the gacela listeners from a project that uses a library on top of gacela. 

> For example: `phel-snake` --(which uses)-> `phel-lang` --(which uses)-> `gacela`

Currently from `phel-snake` you are not able to alter any `gacela` configuration. You can play only with the configuration that `phel-lang` exposes to you. Which make sense "by default", but there are certain things that would be good to be able to expose to any level from gacela point of view; eg **registering listeners** for `gacela events`.

## 🔖 Changes

- Allow registering listeners using `gacela.php` configuration file from a project which uses gacela as vendor of a vendor.

### DIFF
--- a/src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigUsingGacelaPhpFileFactory.php
+++ b/src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigUsingGacelaPhpFileFactory.php
@@ -41,6 +41,8 @@ final class GacelaConfigUsingGacelaPhpFileFactory implements GacelaConfigFileFac
         $gacelaConfig = $this->createGacelaConfig();
         $setupGacela = SetupGacela::fromGacelaConfig($gacelaConfig);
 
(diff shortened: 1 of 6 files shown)


src/Framework/Config/GacelaFileConfig/Factory/GacelaConfigUsingGacelaPhpFileFactory.php
```diff
@@ -41,6 +41,8 @@ public function createGacelaFileConfig(): GacelaConfigFileInterface
         $gacelaConfig = $this->createGacelaConfig();
         $setupGacela = SetupGacela::fromGacelaConfig($gacelaConfig);
 
+        $this->setup->overrideEventDispatcher($setupGacela);
```
Chemaclass, line 44: This is the key part of the PR 🧠

JesusValera (approved):

PR #226, opened 2022-11-06
Fix PhpFileCache

Chemaclass:
## 📚 Description

I discover a clear issue with the current implantation for the `PhpFileCache`, because this one was actually slower than the `InMemoryCache`. After some investigation I found that the object holding the `PhpFileCache` was actually being created every time instead of keeping itself cached. I found this thanks to the internal event system we introduced last week.

## 🔖 Changes

- Add more internal events for the `ClassResolver\Cache` scope
- Fix the `PhpFileCache` holding a unique cache across the project, which boost the performance by almost 50%

### Before

Screenshot 2022-11-06 at 18 05 02

### After

Screenshot 2022-11-06 at 18 03 27

And comparing master with this branch:

Screenshot 2022-11-06 at 18 14 41

> **Note**: The performance boost is crazy good ⚡ Thanks @dantleech for PHPBench!

### DIFF
--- a/tests/Integration/Framework/DocBlockResolverAware/DocBlockResolverCustomServicesAwareTest.php
+++ b/tests/Integration/Framework/DocBlockResolverAware/DocBlockResolverCustomServicesAwareTest.php
@@ -6,6 +6,7 @@ namespace GacelaTest\Integration\Framework\DocBlockResolverAware;
 
 use Gacela\Framework\Bootstrap\GacelaConfig;
 use Gacela\Framework\ClassResolver\Cache\CustomServicesPhpCache;
+use Gacela\Framework\Event\GacelaEventInterface;
 use Gacela\Framework\Gacela;
 use GacelaTest\Feature\Util\DirectoryUtil;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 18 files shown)


Tito-Kati (approved):

tests/Integration/Framework/DocBlockResolverAware/DocBlockResolverCustomServicesAwareTest.php
```diff
@@ -15,14 +16,16 @@ final class DocBlockResolverCustomServicesAwareTest extends TestCase
...
     }
 
     protected function setUp(): void
     {
         Gacela::bootstrap(__DIR__, static function (GacelaConfig $config): void {
             $config->resetInMemoryCache();
             $config->addAppConfig('config/custom-services/*.php');
+            $config->registerGenericListener(static function (GacelaEventInterface $event): void {
+                // dump('Triggered -> ' . \get_class($event)); # useful for debugging
+            });
```
Chemaclass, line 28: This helped me a lot debugging the behaviour by rendering all events that were triggered. Especially the new ones related to the "class name cache" scope.

JesusValera (approved):

PR #229, opened 2022-11-12
Add container factory method

Chemaclass:
## 📚 Description

Currently, the `Container` caches in memory the result of the service after `Container::get('service-key')` is called. This is useful for `Facades` but it might not be the best when creating other types of dependencies.

With this new feature, you can define in your `DependencyProvider` to resolve a particular service every time, avoiding caching it in memory. This is useful when you want to force the creation of a dependency instead of sharing it as singleton.

## 🔖 Changes

- Add `Container::factory()` to avoid caching the result of the service when calling `Container::get('service-key')`.


tests/Unit/Framework/Container/ContainerTest.php
```diff
@@ -64,17 +66,58 @@ public function test_resolve_service_as_function(): void
...
+            $this->container->get('service_name'),
+            $this->container->get('service_name')
+        );
+    }
+
+    public function test_resolve_factory_service_with_random(): void
+    {
+        $this->container->set(
+            'service_name',
+            $this->container->factory(
```
JesusValera, line 103: I am not completely convinced of `factory` name.
Probably, I would rename the methods to: 
- `instance` (for the `factory` method)
- `singleton` (for the `get` method)

What do you think? 🤔

Chemaclass: Good observation, but actually this is the same naming Spryker is using in their [internal Container](https://github.com/spryker/container/blob/master/src/Spryker/Service/Container/Container.php#L587) as well. I kinda like it, and I would prefer to have a similar name in this area.

PR #230, opened 2022-11-12
Change to protected AbstractClassResolver->doResolve()

JesusValera:
## 📚 Description

The `AbstractClassResolver->doResolve()` method is exposed as `public`, but it should be `protected` instead.


Chemaclass (approved):
Good catch, thank you!

PR #235, opened 2022-11-17
Remove $raw array from container

JesusValera:
## 📚 Description

The `$raw` array from `Container` is not necessary, removed 😄


Chemaclass (approved):
Good catch. Less is more!

PR #236, opened 2022-11-18
Add Container::protect()

Chemaclass:
## 📚 Description

Currently, the `DependenyProvider` can define only services that will be resolved on the `Factory`. There is no way to define a service that won't be resolved for any reason.

This PR is about allowing the creation of a service(callable) that won't be resolved when it's been used.

## 🔖 Changes

- A protected service wont be resolved when it's used
- A protected service cannot be extended

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### Unreleased
+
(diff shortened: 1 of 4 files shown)


JesusValera (approved):
Nice PR 😀

PS: Don't forget about the doc.

PR #243, opened 2022-12-04
Locator should not resolve interfaces magically

JesusValera:
## 📚 Description

Similar to the [comment](https://github.com/gacela-project/gacela/pull/203#pullrequestreview-1129816016) I wrote at [Allow using Interface suffix for DocBlockResolver](https://github.com/gacela-project/gacela/pull/203).
We should not depend on the class name to resolve a class from its interface.

This PR aims to remove the logic from the locator that adds some magic to the code and makes it less explicit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Unreleased
 
 - Allow extending raw arrays as services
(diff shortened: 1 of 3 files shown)


Chemaclass (approved):
Fair enough, less is more in this case. Thank you!

PR #250, opened 2023-02-25
Install composer-run-parallel (dev)

Chemaclass:
## 📚 Description

This tool allow us to run composer scripts in parallel. 

> **Note**: More about it here: https://veewee.github.io/blog/run-composer-tasks-in-parallel/

## 🔖 Changes

- Install --dev `veewee/composer-run-parallel`

### Benchmark test
Running `composer ctal` [clear cache and run test-all]
- BEFORE: 30 sec
- AFTER: 20 sec


JesusValera (approved):
It sounds interesting :D

PR #251, opened 2023-03-17
Allow using static facade

Chemaclass:
## 📚 Description

Allow using static method from your Facade directly.

### DIFF
--- a/src/Framework/FactoryResolverAwareTrait.php
+++ b/src/Framework/FactoryResolverAwareTrait.php
@@ -5,10 +5,23 @@ declare(strict_types=1);
 namespace Gacela\Framework;
 
 use Gacela\Framework\ClassResolver\Factory\FactoryResolver;
+use RuntimeException;
 
+/**
+ * @method static AbstractFactory factory()
+ */
 trait FactoryResolverAwareTrait
 {
     private ?AbstractFactory $factory = null;
+
+    public static function __callStatic(string $name, array $arguments): AbstractFactory
+    {
+        if ($name === 'factory') {
+            return (new static())->getFactory();
+        }
+
+        throw new RuntimeException('Unknown method: ' . $name);
+    }
 
     protected function getFactory(): AbstractFactory
     {
--- a/tests/Feature/Framework/StaticFacade/FeatureTest.php
+++ b/tests/Feature/Framework/StaticFacade/FeatureTest.php
@@ -15,6 +15,13 @@ final class FeatureTest extends TestCase
         Gacela::bootstrap(__DIR__);
     }
 
+    public function test_object_facade(): void
+    {
+        $greet = (new TestStaticFacade())->formalGreet('Jesus');
+
+        self::assertSame('Hello, Jesus.', $greet);
+    }
+
     public function test_non_existing_facade_method(): void
     {
         $this->expectExceptionMessage('Unknown method: unknownGreet');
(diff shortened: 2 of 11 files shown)


tests/Feature/Framework/StaticFacade/FeatureTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+        TestStaticFacade::unknownGreet('anything');
+    }
+
+    public function test_static_facade(): void
+    {
+        $greet = TestStaticFacade::informalGreet('Chema');
+
+        self::assertSame('Hi, Chema!', $greet);
+    }
```
Chemaclass, line 37: This is basically the new functionality that I am introducing in this PR 🧠

Tito-Kati (approved):
You can eat on top 👌🏻

src/Framework/FactoryResolverAwareTrait.php
```diff
@@ -5,11 +5,24 @@
...
+use RuntimeException;
 
+/**
+ * @method static AbstractFactory factory()
+ */
 trait FactoryResolverAwareTrait
 {
     private ?AbstractFactory $factory = null;
 
+    public static function __callStatic(string $name, array $arguments): AbstractFactory
```
JesusValera, line 17: Does it make sense to call `parent::__callStatic($name, $arguments)` before doing our own logic?

tests/Feature/Framework/StaticFacade/FeatureTest.php
```diff
@@ -0,0 +1,47 @@
...
+    public function setUp(): void
+    {
+        Gacela::bootstrap(__DIR__);
+    }
+
+    public function test_object_facade(): void
+    {
+        $greet = (new TestStaticFacade())->formalGreet('Jesus');
+
+        self::assertSame('Hello, Jesus.', $greet);
```
JesusValera, line 22: 👋🏼

JesusValera (commented):
I like the idea of having this static call, but my question: In which scenarios someone would prefer to call a static method instead of a non-static method? 🤔

Chemaclass:
>  In which scenarios someone would prefer to call a static method instead of a non-static method?

@JesusValera, good question! I am writing some new examples from scratch and I felt it was simpler to access the facade using a static method. eg:
```php
$isSpam = CommentFacade::isSpam('Lorem ipsum!');
# vs
$isSpam = (new CommentFacade())->isSpam('Lorem ipsum!');
```
In the end, it's up to you, the dev, how to use it. Right now using static calls require some boilerplate and extra complexity for the developer. So the idea is to simplify and allow using the static syntax if you want to as well.

JesusValera (approved):

src/Framework/FactoryResolverAwareTrait.php
Chemaclass: wow! I didn't think about this 🤔
I don't know, actually, do you see any potential problem if we don't have this parent call?

JesusValera: It pops to my mind because we had already some problems with Laravel in the past, the magic methods like `__call`, `__callStatic` and so on are special methods, and it's easier to get conflicts with other frameworks/user-defined classes if we don't take that into account... 👀 
But cannot sentence anything at the moment.

We can try as it is for now, and if someone complains in the future or we spot some issue, we can change it 👍🏼

PR #252, opened 2023-03-18
ResetInMemoryCache also from anonymous globals and factory containers

Chemaclass:
## 📚 Description

When using `resetInMemoryCache()` we should remove the internal cached state from Gacela; this is really useful for feature/integration tests. Eg:

```php
    $config->resetInMemoryCache();
});
```

## 🔖 Changes

- `ResetInMemoryCache` should work also from "anonymous globals" and "factory containers"

### DIFF
--- a/src/Framework/ClassResolver/GlobalInstance/AnonymousGlobal.php
+++ b/src/Framework/ClassResolver/GlobalInstance/AnonymousGlobal.php
@@ -21,6 +21,14 @@ final class AnonymousGlobal
     private static array $cachedGlobalInstances = [];
 
     /**
+     * @internal
+     */
+    public static function resetCache(): void
+    {
+        self::$cachedGlobalInstances = [];
+    }
+
+    /**
      * @template T
      *
      * @param class-string<T> $className
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Unreleased
 
 - Allow using static facade methods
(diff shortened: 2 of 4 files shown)


src/Framework/ClassResolver/GlobalInstance/AnonymousGlobal.php
```diff
@@ -20,6 +20,14 @@ final class AnonymousGlobal
     /** @var array<string,object> */
     private static array $cachedGlobalInstances = [];
 
+    /**
+     * @internal
+     */
+    public static function resetCache(): void
```
JesusValera, line 26: The method is marked with `@ignore`, even though, I would add a comment explaining why this method is public and what to do if the client want to reset the cache, I mean, which other alternatives exists (if it makes sense) and why is discouraged to call this method directly from somewhere else outside Gacela.
I think there are also a few public methods with `@ignore` keyword, same for them, what do you think?

Chemaclass: I thought/think the function name is self explanatory enough 🤔
As you commented, there are multiple other similar functions for the other resetCache parts. 
Screenshot 2023-03-18 at 13 44 09

---

If any dev would try to use an internal method, it's by their own responsibility. They will get a warning from psalm/phpstan telling them that the method is internal, so they should not use it.
Screenshot 2023-03-18 at 13 47 32

JesusValera (approved):

PR #253, opened 2023-03-18
Allow using ::config() from any Factory

Chemaclass:
## 📚 Description

Inspired by https://github.com/gacela-project/gacela/pull/251 

## 🔖 Changes

- Allow using static `::config()` from any Factory

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ### Unreleased
 
 - Allow using static facade methods
+  - Allow using static `::factory()` from any Facade
+  - Allow using static `::config()` from any Factory
 - ResetInMemoryCache also from anonymous globals and factory containers
 
 ### 1.0.1
--- a/src/Framework/ConfigResolverAwareTrait.php
+++ b/src/Framework/ConfigResolverAwareTrait.php
@@ -11,6 +11,14 @@ trait ConfigResolverAwareTrait
 {
     private ?AbstractConfig $config = null;
 
+    /**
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ### Unreleased
 
 - Allow using static facade methods
+  - Allow using static `::factory()` from any Facade
+  - Allow using static `::config()` from any Factory
```
Chemaclass, line 7: This is the new one: `::config()`.
>**Note**: "Allow using static `::factory()` from any Facade" was already implemented here https://github.com/gacela-project/gacela/pull/251

src/Framework/ConfigResolverAwareTrait.php
```diff
@@ -11,6 +11,14 @@ trait ConfigResolverAwareTrait
...
 
+    /**
+     * Syntax sugar to access the config from static methods.
+     */
+    public static function config(): AbstractConfig
+    {
+        return (new static())->getConfig();
+    }
+
     public function getConfig(): AbstractConfig
```
JesusValera, line 22: I am not sure about naming the method `config()` when it is static and `getConfig()` when it is not.
Maybe it would be better to unify the name in both situations?

Chemaclass: The thing is that you cannot have the same func name for both (non-static and static function). Unless we use magic methods like __call and __staticCall ... 🤔 

> I will investigate this part. Thanks for bringing this up 🙇🏼

PR #254, opened 2023-03-18
Allow using factory() and getFactory() from Facade

Chemaclass:
## 📚 Description

Follow up from this idea https://github.com/gacela-project/gacela/pull/253#discussion_r1141150157
Screenshot 2023-03-19 at 00 36 48

## 🔖 Changes

- Allow using `factory()` or `getFactory()` from an object or class context

### Test A [static factory()]
Screenshot 2023-03-19 at 00 38 54

### Test B [static getFactory()]
Screenshot 2023-03-19 at 00 39 26

## 💡 Follow up

If this idea gets merged, we can do something similar for the relation between `factory->config()`


JesusValera (approved):
That is exactly what I was thinking, nice! 💃🏼

Chemaclass:
There is a little problem with this approach, though... I think it creates confusion rather than helping, when you see these two methods. As a dev, I would ask myself, "why do I have these two methods?" The fact that they do the same is not a valid excuse... We need to rethink this approach, @JesusValera  🤔 
Screenshot 2023-03-19 at 15 27 31

Chemaclass:
Closing the PR because I don't like to add this magic. I think keeping the concrete `$object->getFactory()` is good. And the optional `self::factory()` when using it from a static context, this is a little convention that we can create, way simpler than allowing both for both contexts.

PR #255, opened 2023-03-19
Add git-hooks when running "composer install" automatically

JesusValera:
## 📚 Description

- Move the `tools` dir to `.github`
- Add `post-install-cmd` script to install git hooks when updating dependencies


Chemaclass (approved):
I like this idea 💡

PR #258, opened 2023-04-10
Move DependencyResolver to a separate library

Chemaclass:
## 📚 Description

Move the logic from `InstanceCreator` and `DependencyResolver` to a new and decoupled library: 
`gacela-project/dependency-resolver` ([packagist](https://packagist.org/packages/gacela-project/dependency-resolver))


composer.json
```diff
@@ -27,7 +27,8 @@
     "prefer-stable": true,
     "minimum-stability": "dev",
     "require": {
-        "php": "^8.0, <8.3"
+        "php": "^8.0, <8.3",
+        "gacela-project/dependency-resolver": "^0.1"
```
JesusValera, line 31: Thinking about the package name... 🤔 

Some proposals:
- `Dependency Resolver`
- `Dependency Container`
- `Dependency Injector`
- `Service Container` 
- `Container` -> [Laravel](https://laravel.com/docs/10.x/container)
- `Dependency Injection` -> [Symfony](https://symfony.com/doc/current/components/dependency_injection.html)
- `Object Resolver` (was the name proposed by @Chemaclass?)
- you say ...!

I think I like `Dependency Resolver`, but open to discuss.
But I think we should add these proposals to the `keywords` key on L13 to have more visibility, what do you think?

JesusValera (approved):

composer.json
Tito-Kati: Resolver ☝🏻


Another option would be Binder.

Chemaclass: I like that. I renamed it to just "resolver" -> `gacela-project/resolver`

Tito-Kati (approved):
👌️

PR #263, opened 2023-04-25
Allow running code right after bootstraping

Chemaclass:
## 📚 Description

There is no way to run custom code right after calling `Gacela::bootstrap()`. 

The goal is to allow running custom code that doesn't need to be define in the entry `index.php`, mainly to set a more complex custom configuration for the dedicated client app.

> Idea inspired by https://github.com/gacela-project/router/issues/25

## 🔖 Changes

- Create a `PluginInterface` which can be custom logic 
- And you can add your custom class in the `GacelaConfig::prePlugings()`
  - Note: it has to extends that new `PluginInterface`

## 🧪  Example

```php
    $config->prePlugins([
      YourCustomLogicPlugin::class,
    ]);
});
```

>**Note**: The `YourCustomLogicPlugin` will be instantiated with the gacela container, which means we got auto-wiring for its dependencies/collaborators.

#### Using the `Container->getLocator()->get(class)`

This img is just an example of another app using this new feature and showing its potential and usage:

Screenshot 2023-04-25 at 16 21 41

#### Using the global `Gacela::get(class)`

Another option would be to use `Gacela::get(class)` directly there:

Screenshot 2023-04-25 at 16 28 06

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 - Unify `setFileCacheEnabled` and `setFileCacheDirectory` into one single method: `setFileCache(bool $enabled, string $dir)`. Deprecated the former methods.
 - Rename dependency; from `resolver` to `container`.
-- Moved the current `Container` logic to the decoupled `container` dependency. 
+- Moved the current `Container` logic to the decoupled `container` dependency.
+- Add "prePlugins" to run right after the `Gacela::bootstrap()`
 
 ### 1.1.1
 ### 2023-04-19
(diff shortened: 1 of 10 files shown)


Chemaclass:
@JesusValera @Tito-Kati I updated the screenshot in the PR description, and some ideas behind:
The Locator is an internal class, so the user shouldn't be dealing with it directly. Instead, they should use it indirectly via other methods, so we can keep control of what is allowed; eg: `Gacela::get()` or `Container->getLocator()->get()`.

Tito-Kati (approved):
Awesome work!

JesusValera (approved):

Issue #266, opened 2023-04-26
Rename mappingInterface to bind


Chemaclass:
## 📚  Description

We use a lot of usages for `mappingInterfaces(interface, concrete)` in the project. 
I think it would be simpler to use a `bind(key, value)` concept; similar as used in the [container project](https://github.com/gacela-project/container/blob/main/src/Container/Container.php#L40)

PR #268, opened 2023-04-28
Use main container on plugins

Chemaclass:
## 📚 Description

I discovered a bug while developing a `api-skeleton` repository, where I am using all gacela components together:
 
Screenshot 2023-04-28 at 20 17 22

The reason was that after the change https://github.com/gacela-project/router/pull/28 we removed the three collaborator classes (via constructor) and instead we are passing an optional Clojure to the Router class. This, however, wasn't working fine with our current gacela container because it didn't know how to resolve those dependencies.

The solution was to allow defining external global bindings (currently known as `mappingInterfaces` at `GacelaConfig` level (extra: there is an issue to rename this concept to bindings https://github.com/gacela-project/gacela/issues/266 as done in the [gacela container project](https://github.com/gacela-project/container/blob/main/src/Container/Container.php#L40)).

## 🔖 Changes

- Pass down the main container to the Locator singleton created when doing `Gacela::bootstrap()`.
  - This is mandatory step to setup Gacela in your project.
  - This allow you to setup global singletons that you can access later on at any time using `Gacela::get(key)` which is useful, for example, if you want to define a single instance of a Router and you want to alter it to add different routes depending on each module.

### DIFF
--- a/src/Framework/Container/Container.php
+++ b/src/Framework/Container/Container.php
@@ -22,6 +22,6 @@ final class Container extends GacelaContainer implements ContainerInterface
 
     public function getLocator(): LocatorInterface
     {
-        return Locator::getInstance();
+        return Locator::getInstance($this);
     }
 }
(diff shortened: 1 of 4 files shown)


Tito-Kati (approved):
Nice bug hunting!

JesusValera (approved):

PR #270, opened 2023-05-03
Install infection

Chemaclass:
## 📚 Description

Simila as we did on container and router, we want to install and get the benefits from infection php mutation testing framework on the core repository of gacela.

## 🔖 Changes

- install infection

## 🖼️  Status quo

Screenshot 2023-05-03 at 12 04 14

### DIFF
--- a/.github/workflows/mutation-testing.yml
+++ b/.github/workflows/mutation-testing.yml
@@ -1,0 +1,42 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+name: Mutation Tests
+
+jobs:
+
+  mutation-tests:
+    name: Mutation Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: 8.0
+          coverage: xdebug
+          tools: composer
+
+      - name: Get composer cache directory
+        id: composer-cache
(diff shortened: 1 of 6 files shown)


.github/workflows/mutation-testing.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+
+      - name: Cache dependencies
+        uses: actions/cache@v3.3.1
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
```
localheinz, line 33: @Chemaclass 

`composer.lock` is currently not checked in and does not exist, so `actions/cache` can not assemble a cache key based on `composer.lock`.

Perhaps change this to

```suggestion
          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
```

or check in `composer.lock` - also helps with reproducible builds.

What do you think?

Chemaclass: You're right. However, we are using the same line in the other workflows ([code-style](https://github.com/gacela-project/gacela/blob/main/.github/workflows/code-style.yml#L33), for example), and no issues there

Chemaclass:
I am using the same as in the docs https://github.com/shivammathur/setup-php#cache-composer-dependencies but still something seems to be wrong... any more ideas, @localheinz ?

Chemaclass:
I will merge this as it adds infection to the project. The problem with "composer caching" is another thing that we can postpone to another time.

PR #271, opened 2023-05-06
Rename PluginInterface main method

Chemaclass:
## 📚 Description

The current `PluginInterface` forces you to implement that interface if you want to create a plugin, which forces you to require the entire gacela package as dependency.

However, it would be useful to allow any invokable/callable thing to be executed as a plugin, so you can define plugins that will be instantiated later without the need of depending on the fully gacela as dependency.

## 🔖 Changes

- Rename the `PluginInterface::run()` as just `PluginInterface::__invoke()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Unreleased
+
+- Rename PluginInterface main method from `run()` to `__invoke()` 
+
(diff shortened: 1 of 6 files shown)


Tito-Kati (approved):
Less coupling! Less coupling! 🤩️

JesusValera (approved):

PR #274, opened 2023-05-08
Add extend config

Chemaclass:
## 📚 Description

Follow up after https://github.com/gacela-project/gacela/pull/272 

The idea is to allow extending the `GacelaConfig` from a different scope before finalising gacela's bootstrapping. 

> This is pretty handy because it allows you to configure/bootstrap gacela from different places, not just from the `Gacela::bootstrap()` function or `gacela.php` file. 
> 
> Eg: you can define a custom config -like binding an interface or final class to a concrete instance- from your own library that later will be used by other client.

## 🔖 Changes

- Add "extendConfig" method to the `GacelaConfig` object.
- Rollback https://github.com/gacela-project/gacela/pull/272 
  - Now that there is no need to have a before/after concept for plugins, we can use "plugins" to keep it simple


JesusValera (approved):
Better naming!

Tito-Kati (approved):
Cool!

PR #276, opened 2023-05-14
Add Gacela::rootDir()

Chemaclass:
## 📚 Description

I think it would be useful to be able to access to the rootDir of the application anywhere is needed, especially for configuration stuff.

>**Note** Thanks @JesusValera for the idea 💡 

## 🔖 Changes

- Save the appRootDir as static prop in the Gacela class and make it accesible from anywhere as static method.

### DIFF
--- a/tests/Integration/Framework/GacelaTest.php
+++ b/tests/Integration/Framework/GacelaTest.php
@@ -7,11 +7,19 @@ namespace GacelaTest\Integration\Framework;
 use Gacela\Framework\Bootstrap\GacelaConfig;
 use Gacela\Framework\Gacela;
 use PHPUnit\Framework\TestCase;
+use ReflectionClass;
 
 final class GacelaTest extends TestCase
 {
     public function test_null_get_cache_dir(): void
     {
+        // Needed in order to set up the private `Gacela::$appRootDir as null` for this use case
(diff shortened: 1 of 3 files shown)


tests/Integration/Framework/GacelaTest.php
```diff
@@ -0,0 +1,37 @@
...
+    public function test_null_get_cache_dir(): void
+    {
+        // Needed in order to set up the private `Gacela::$appRootDir as null` for this use case
+        // when you try to access it before bootstrapping the application.
+        $reflectionClass = new ReflectionClass(Gacela::class);
+        $reflectionProperty = $reflectionClass->getProperty('appRootDir');
+        $reflectionProperty->setAccessible(true);
+        $reflectionProperty->setValue(null);
+
+        self::assertNull(Gacela::rootDir());
```
JesusValera, line 23: Can't you do the following test?
```php
# test A
self::assertNull(Gacela::rootDir());

# test B
Gacela::bootstrap(__DIR__);

self::assertSame(__DIR__, Gacela::rootDir());
```

Chemaclass: This is exactly what I did in these two tests. Or what do you mean?

PR #284, opened 2023-05-28
Add command list modules

Chemaclass:
## 📚 Description

It would be useful to be able to display a list with all existing modules starting from the app root directory.

## 🔖 Changes

- Create a new console command to list all modules based on the fact that if there is a class that extends the Gacela `AbstractFacade` then that's a valid "Gacela module"

## 🖼️ Screenshots

An example of running ` php gacela list:modules` on the gacela project: it searches for the modules inside the search directory. As a follow up, we could add an additional argument to the command to filter for that specific string. 

Screenshot 2023-05-28 at 20 21 16


Tito-Kati (approved):
👌🏼🔝

PR #285, opened 2023-05-30
Filter list modules command

JesusValera:
## 📚 Description

Add the `filter` argument to `gacela list:module` command:

image

### DIFF
--- a/src/Console/ConsoleFacade.php
+++ b/src/Console/ConsoleFacade.php
@@ -40,10 +40,10 @@ final class ConsoleFacade extends AbstractFacade
     /**
      * @return list<AppModule>
      */
-    public function findAllAppModules(): array
+    public function findAllAppModules(string $filter): array
     {
         return $this->getFactory()
             ->createAllAppModulesFinder()
-            ->findAllAppModules();
+            ->findAllAppModules($filter);
     }
 }
--- a/src/Console/ConsoleFacade.php
+++ b/src/Console/ConsoleFacade.php
@@ -40,7 +40,7 @@ final class ConsoleFacade extends AbstractFacade
     /**
      * @return list<AppModule>
      */
-    public function findAllAppModules(string $filter): array
+    public function findAllAppModules(?string $filter): array
     {
         return $this->getFactory()
             ->createAllAppModulesFinder()
--- a/src/Console/ConsoleFacade.php
+++ b/src/Console/ConsoleFacade.php
@@ -40,7 +40,7 @@ final class ConsoleFacade extends AbstractFacade
     /**
      * @return list<AppModule>
      */
-    public function findAllAppModules(?string $filter): array
+    public function findAllAppModules(string $filter = ''): array
     {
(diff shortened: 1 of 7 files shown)


src/Console/Infrastructure/Command/ListModulesCommand.php
```diff
@@ -21,12 +22,14 @@ final class ListModulesCommand extends Command
     protected function configure(): void
     {
         $this->setName('list:modules')
-            ->setDescription('Render all modules found');
+            ->setDescription('Render all modules found')
+            ->addArgument('filter', InputArgument::OPTIONAL, 'The filter path. For example "App/TestModule/TestSubModule');
```
Chemaclass, line 26: I would suggest this description to something simpler:
```suggestion
            ->addArgument('filter', InputArgument::OPTIONAL, 'Any filter to simplify the output');
```
What do you think, @JesusValera and @Tito-Kati ?

src/Console/Domain/AllAppModules/AllAppModulesFinder.php
```diff
@@ -35,17 +35,25 @@ public function findAllAppModules(): array
...
             || $fileInfo->getExtension() !== 'php'
             || str_contains($fileInfo->getRealPath(), 'vendor/')
         ) {
             return null;
         }
+
         $namespace = $this->getNamespace($fileInfo);
         $className = $this->buildClassName($fileInfo);
 
+        if (!empty($filter)) {
```
Chemaclass, line 50:
```suggestion
        if ($filter !== '') {
```

tests/Feature/Console/ListModules/FilterListModulesCommandTest.php
```diff
@@ -0,0 +1,34 @@
...
+use Symfony\Component\Console\Input\StringInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+
+final class FilterListModulesCommandTest extends TestCase
+{
+    public function test_list_modules_with_filter(): void
+    {
+        Gacela::bootstrap(__DIR__);
+
+        $input = new StringInput('list:modules GacelaTest/Feature/Console/ListModules/TestModule1');
```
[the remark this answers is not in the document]
Chemaclass: Would it make sense to allow also `list:modules ListModules\\TestModule1`?
I think I would like to allow both options (this one does not work with the current logic, so you need to implement it) 🤔

src/Console/ConsoleFacade.php
```diff
@@ -40,10 +40,10 @@ public function generateFileContent(
     /**
      * @return list<AppModule>
      */
-    public function findAllAppModules(): array
+    public function findAllAppModules(?string $filter): array
```
Chemaclass, line 43: What about this?
```suggestion
    public function findAllAppModules(?string $filter = null): array
```

Chemaclass (approved):
Awesome. Thank you! 🙏

PR #286, opened 2023-05-31
Fix: Replace `locked` with `highest` in matrix

localheinz:
## 📚 Description

This pull request

- [x] replaces `locked` with `highest` in the matrix

💁‍♂️ `composer.lock` is currently not checked in, so there is no way to install locked dependencies.

## 🔖 Changes

- List individual changes in more detail

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        dependencies: [ "locked", "highest" ]
+        dependencies: [ "lowest", "highest" ]
         php-version: [ "8.0" , "8.1", "8.2" ]
         operating-system:
           - "ubuntu-latest"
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -37,17 +37,13 @@ jobs:
         uses: actions/cache@v3.3.1
         with:
           path: ${{ steps.composercache.outputs.dir }}
-          key: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}-${{ hashFiles('**/composer.lock') }}"
-          restore-keys: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}-${{ hashFiles('**/composer.lock') }}"
```
localheinz, line 41: `actions/cache` automatically use the value for `key` as a restore key. The `restore-keys` option allows to specify additional fallback keys.

Chemaclass (approved):
Thanks a lot!

PR #297, opened 2023-06-25
Use detailed by default instead of simple

Chemaclass:
## 📚 Description

Original idea from @JesusValera.

## 🔖 Changes

- Use the detailed view of the command by default. Keep the simplified view (displaying only the module names) by using an optional option `--simple|-s`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Unreleased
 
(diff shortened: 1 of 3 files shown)


JesusValera (approved):
I have some ideas, let me think of them in a forward PR 😄

Chemaclass:
@JesusValera yes, feel free to create a follow up PR with your ideas 🧠

PR #298, opened 2023-06-26
Change design simple view from gacela list:modules

JesusValera:
## 📚 Description

Change design from the simple view:

image

to the following table:

image

### DIFF
--- a/src/Console/Domain/AllAppModules/AppModuleCreator.php
+++ b/src/Console/Domain/AllAppModules/AppModuleCreator.php
@@ -12,6 +12,8 @@ use Gacela\Framework\ClassResolver\Factory\FactoryNotFoundException;
 use Gacela\Framework\ClassResolver\Factory\FactoryResolver;
 use ReflectionClass;
 
+use function strlen;
+
 final class AppModuleCreator
 {
     public function __construct(
@@ -27,6 +29,7 @@ final class AppModuleCreator
     public function fromClass(string $facadeClass): AppModule
     {
         return new AppModule(
+            $this->fullModuleName($facadeClass),
             $this->moduleName($facadeClass),
             $facadeClass,
             $this->findFactory($facadeClass),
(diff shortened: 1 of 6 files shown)


khru (commented):
Love the emojis

src/Console/Domain/AllAppModules/AppModuleCreator.php
```diff
@@ -35,15 +38,26 @@ public function fromClass(string $facadeClass): AppModule
...
     /**
      * @param class-string $facadeClass
      */
     private function moduleName(string $facadeClass): string
     {
-        $parts = explode('\\', $facadeClass);
-        array_pop($parts);
+        $fullModuleName = $this->fullModuleName($facadeClass);
+
+        $moduleName = strrchr($fullModuleName, '\\') ?: $fullModuleName;
```
Chemaclass, line 58: `strrchr` rrrrrrrrrr 😄

src/Console/Infrastructure/Command/ListModulesCommand.php
```diff
@@ -25,55 +26,62 @@ protected function configure(): void
...
     {
-        return ($isSimple)
-            ? $this->generateSimpleView($modules)
-            : $this->generateDetailedView($modules);
+        if ($isDetailed) {
+            $this->generateDetailedView($output, $modules);
+            return;
+        }
+
+        $this->generateSimpleView($output, $modules);
```
Chemaclass, line 55: What about doing this instead?
```php
  ($isDetailed)
    ? $this->generateDetailedView($modules)
    : $this->generateSimpleView($modules);
```

src/Console/Infrastructure/Command/ListModulesCommand.php
```diff
@@ -25,55 +26,62 @@ protected function configure(): void
...
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $filter = (string)$input->getArgument('filter');
 
-        $listOfModules = $this->generateListOfModules(
-            (bool)$input->getOption('simple'),
+        $this->generateListOfModules(
+            $output,
```
Chemaclass, line 37: What about setting the `output` as private property of the command (eg: in the like 34), so you can access it any level below (on any private func) instead of passing it as an extra argument to 2 or 3 nested methods?

src/Console/Infrastructure/Command/ListModulesCommand.php
```diff
@@ -25,55 +26,62 @@ protected function configure(): void
...
-            $result .= <<<TXT
-{$n}.- <fg=green>{$module->moduleName()}</>
-
-TXT;
+        $rows = [];
+
+        foreach ($modules as $module) {
+            $rows[] = [
+                $module->fullModuleName(),
+                '✔️', // facade is always true
```
Chemaclass, line 68: What about using a const for these emojis? Not mandatory, just wondering what would you think 🤔

Tito-Kati:
Awesome!! 🙌🏻🙌🏻🙌🏻

Tito-Kati (approved):

src/Console/Infrastructure/Command/ListModulesCommand.php
JesusValera: I think `else` in this scenario fits better, it is more explicit than your suggestion 😄

src/Console/Infrastructure/Command/ListModulesCommand.php
JesusValera: Agree, good idea! 💡

Chemaclass (approved):
Genial 👌🏼

PR #299, opened 2023-07-09
Refactor FactoryResolver

Chemaclass:
## 📚 Description

There is no need to have a `FactoryNotFoundException` because the `AbstractClassResolver` will resolve always a Factory even if there is none in a module, creating always a default anonymous class which extends the `AbstractFactory` (see: `createDefaultGacelaClass()`)

## 🔖 Changes

- Simplify `FactoryResolver`
- Remove `FactoryNotFoundException`

## 🤔  TODO

We should do the same to the `ConfigResolver` in a follow up PR.


JesusValera (approved):
👌🏼

fabriziofs (approved):
Nice! 👌

PR #318, opened 2024-08-11
Enable using addGlobal and overrideExistingResolvedClass from Gacela

Chemaclass:
## 📚 Description

Replace this text with a short description of your feature/bugfix. 

## 🔖 Changes

- Add `addGlobal(object $resolvedClass, object|string $context = '')`
  - Make `$context` optional, using the caller's file as context if none is provided

- Add `overrideExistingResolvedClass(string $className, object $resolvedClass)`


src/Framework/Gacela.php
```diff
@@ -78,6 +78,26 @@ public static function rootDir(): string
         return self::$appRootDir;
     }
 
+    /**
+     * Add an anonymous class as 'Config', 'Factory' or 'DependencyProvider' as a global resource
+     * bound to the context that it is passed as first argument.
```
JesusValera, line 83: Do you mean the `second argument`?

src/Framework/Gacela.php
```diff
@@ -78,6 +78,26 @@ public static function rootDir(): string
...
+     * It can be the string-key (from a non-class/file context) or the class/object itself.
+     */
+    public static function addGlobal(object $resolvedClass, object|string $context = ''): void
```
JesusValera, line 86: Question: What if you add two resolvers for the same `context` instead of calling `overrideExistingResolvedClass()` method?

JesusValera (approved):

src/Framework/Gacela.php
Chemaclass: @JesusValera, it would not work. Let's try changing the test here: [AnonymousGlobalTest.php#L62](https://github.com/gacela-project/gacela/blob/main/tests/Unit/Framework/ClassResolver/GlobalInstance/AnonymousGlobalTest.php#L62) 

Screenshot 2024-08-11 at 22 18 00

> OUTCOME: RuntimeException: Type 'App\Module\ClassNameFacade' not allowed. Valid types: Config, Factory, DependencyProvider

### Reason

- On `addGlobal()` the context must be a file path or an object (usually `$this`)
- However, `overrideExistingResolvedClass()` is to refer an existing gacela class, being able to overriding it (eg: `YourCustomDependencyProvider::class` for testability reasons, or any other)

PR #319, opened 2024-08-17
Add AbstractProvider

Chemaclass:
## 🔖 Changes

- Deprecate `AbstractDependencyProvider` in favor of `AbstractProvider`


src/Framework/AbstractFacade.php
```diff
@@ -4,7 +4,54 @@
...
+    /** @var array<string, AbstractFactory> */
+    private static array $factories = [];
+
+    /**
+     * @param list<mixed> $arguments
+     */
+    public static function __callStatic(string $name = '', array $arguments = [])
+    {
+        if ($name === 'getFactory') {
+            return self::doGetFactory();
```
JesusValera, line 29: Should we call the parent call and staticCall in case they exists? Does it make sense?

khru (approved):

JesusValera (approved):

src/Framework/AbstractFacade.php
Chemaclass: I don't understand your question. For some context, check the tests here: `tests/Feature/Framework/StaticFacade/FeatureTest.php`

Screenshot 2024-08-17 at 17 10 11

PR #447, opened 2026-07-18
Raise mutation score from 87% to 99% MSI

JesusValeraDev:
## 📚 Description

Raises the mutation score from 87% to **99% MSI** (1668 mutants: 1615 killed, 4 timed out, 0 not covered, 7 escaped, 40 documented equivalent-mutant ignores), and fixes a real bug the work surfaced: contextual bindings (`$config->when(X)->needs(Y)->give(Z)`) were silently ignored when resolving Gacela classes, because the class resolver passed a `\`-prefixed class name to the container that never matched the `X::class`-keyed bindings — the global binding always won.

## 🔖 Changes

- **Fix**: normalize the resolved class name in `AbstractClassResolver::createInstance()` so contextual bindings apply to factories/configs/providers; covered by a new `ContextualBindingsResolution` feature test
- Add ~150 test cases killing escaped mutants across:
  - `FileCache` (batch commit + lock file, TTL boundary at `expiresAt === now`, stats aggregation, atomic-write failure paths, path normalization), `ScopedCache` (graph persistence after invalidation, BFS past already-seen nodes, crafted cyclic graph files), `WritableDirectory`
  - `ContainerFixture` (subclass fixture pinning `protected` visibility, snapshot capture/restore, temp-dir naming), plus event getters, `InMemoryCache::getAll()`, `PhpConfigReader`, `DocBlockServiceResolver` defaults, `DocBlockResolver` failure paths
  - All PHPStan rules: richer good/bad fixtures (dynamic method names, non-`$this` targets, `cached()` edge cases, static property fetches, dedup of distinct violations, every skip-branch ordering) and a default-`modulePathSegments` check
  - `CacheableTrait`, `ProvidesScanner` (untyped/scalar params), `SetupMerger` (handler registries + lazy services), `HealthCheckRegistry` (container-provided instances, unresolvable class strings), `DoctorCommand` (degraded/unhealthy metadata rendering), `Config` cache-dir edge cases, `SetupGacela::fromFile()`, `Gacela::getRequired()`
- Stop excluding `src/Framework/Event/*` from the phpunit coverage source — their mutants could never be marked as covered
- Drop stale `@codeCoverageIgnore` annotations from now-tested methods
- Document per-mutator, method-scoped ignores in `infection.json5` for analyzed-equivalent mutants (Windows-only `PHP_OS` branches, `flock()`/opcache side effects unobservable in-process, `isset()`-on-`false` BFS bookkeeping, AST-type guards the PHPStan analyser cannot violate, TOCTOU-only mkdir races)
- Gitignore the `gacela-merged-config.php` debris left behind by mutation runs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,6 @@ tests/**/Users/
 gacela-local.php
 gacela-custom-services.php
 gacela-class-names.php
+gacela-merged-config.php
 .gacelagacela-*.php
 mutation-report.html
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Contextual bindings (`$config->when(X)->needs(Y)->give(Z)`) now apply when resolving Gacela classes (factories, configs, providers): the class resolver passed a `\`-prefixed class name to the container, which never matched the `X::class`-keyed contextual bindings, so the global binding always won
 - Health checks registered in `gacela.php` are no longer wiped when a project also ships `gacela-{APP_ENV}.php`: `GacelaConfig::__construct()` no longer resets the shared `HealthCheckRegistry` on every instantiation. The registry is now cleared exactly once per `Gacela::bootstrap()`, so checks from the default and env config files accumulate together without leaking across bootstraps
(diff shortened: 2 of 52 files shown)


Chemaclass (approved):
let's go!

Issue #539, opened 2026-07-26
ref: one container with an explicit scope seam (2.1)


Chemaclass:
> **Status: blocked on [container#106](https://github.com/gacela-project/container/issues/106), retargeted to 2.1.** The consolidation needs a parent/child container primitive that `gacela-project/container` does not have. See *Where this landed* at the bottom.
>
> **This body has been rewritten.** The original framed the problem as widespread static-cache rot and cited "18 stateful classes, 7 with no reset, 4 declaring a reset nobody calls". The audit in this thread found **one** actual leak. Those numbers are corrected below rather than left standing, because they were the case for doing this and they did not hold.

## Problem

Gacela does not have *a* container. It has a module container plus a spread of process-global static caches, each with its own lifetime and its own reset.

What the audit actually found, of 18 classes in `src/Framework` holding `static array` state:

| | claimed | measured |
|---|---|---|
| declaring a reset `Gacela::resetCache()` never calls | 4 | **2** (`ClassValidator`, `ReflectionClassPool`; `EventDispatcherProvider` is reached transitively, `CacheableConfig` is excluded on purpose) |
| holding state with no reset at all | 7 | **7, none of them broken** — each is a pure function of immutable class facts, or is refreshed on construction |
| actually leaking | — | **1** — `ClassValidator`, memoizing the *negative* `class_exists()` answer. Fixed in #540 |

An eighth was later found by writing the guard rather than by audit: the in-memory copy of the file-backed caches survived `resetCache()` (fixed, PR #557).

## So what is the argument, if the caches are not rotting

Not "static caches are silently rotting everywhere" — measured, they are not. The argument is that **nothing enforces the distinction**.

Whether a given cache needs invalidation is decided per class, by reading it. `ClassValidator` was wrong for a year and its own unit test passed the whole time. `AbstractPhpFileCache` is the counter-example that proves the point: its statics *are* correctly keyed by `static::class`, so subclasses do not collide — that is discipline every other layer has to reproduce by hand, and the list above is where it was not reproduced.

The `CHANGELOG` from 1.14 → 1.20 carries **16** cache-related fixes. They are the same bug wearing different clothes — two things sharing a cache slot, or a clear path missing one layer:

- 1.20 — `provideModuleDependencies()` ran twice: two provider resolvers shared a normalized cache slot
- 1.18 — `ConfigLoader`'s read cache was not keyed by reader, so the same file under two readers shared an entry
[...]

PR #540, opened 2026-07-26
fix(framework): clear the class-exists and reflection caches on resetCache

Chemaclass:
## 📚 Description

`ClassValidator` memoizes `class_exists()` — **including the negative answer**. A class that was not loadable when first asked stays "missing" for the life of the process, and `Gacela::resetCache()` did not clear it.

The reason it was missed is the interesting part: `ClassValidator::resetCache()` and `ReflectionClassPool::reset()` were both implemented, both unit-tested, and both reached **only by the tests written to exercise them**. Production called neither. The suite was green because the tests reached around the public API that real callers use.

Bites any process where the set of loadable classes changes after the first resolution — long-running workers (RoadRunner, Swoole, queue consumers) that re-bootstrap, code generation, `cache:warm` emitting classes. The module resolves to "not found" with nothing pointing at the stale answer.

Found while measuring the ground truth for #539, not by hitting it in the wild.

## 🔖 Changes

- `Gacela::resetCache()` now calls `ClassValidator::resetCache()` and `ReflectionClassPool::reset()`
- `ClassValidatorResetTest` — behavioural: a class that becomes loadable is seen after `resetCache()`. Verified red before the fix (`Failed asserting that false is true`)
- `ResetCacheCoverageTest` — guards the wiring, so the next static cache can't be added without one. A class in `src/Framework` declaring a reset must be reached by `Gacela::resetCache()`, directly or through a reset it calls, or be exempted **with a reason**
- Exemptions are **self-invalidating**: an entry that stops describing reality fails as loudly as a missing reset. Verified both directions — removing the `ClassValidator` call fails the coverage test; adding a `CacheableConfig::reset()` call fails the exemption test

### On the exemptions

Two classes hold static state that must *not* be cleared, and the distinction matters more than the fix:

- **`CacheableConfig`** — holds the cache backend registered by the application via `CacheableConfig::setStorage()`. Nothing in the framework sets it, so clearing it from `resetCache()` would silently drop the app's configured storage.
- **`HealthCheckRegistry`** — holds checks registered through `GacelaConfig::addHealthCheck()`; bootstrap resets it explicitly.

Configuration lifetime is not cache lifetime. Today that distinction lives in a test's allow-list; making it explicit in the container is what #539 is for.

Also corrected an over-claim of mine on #539: `EventDispatcherProvider` is **not** a leak — `Config::resetInstance()` resets it, and `Gacela::resetCache()` calls that. I had counted direct calls only. The coverage test resolves one hop of indirection so it doesn't repeat my mistake.

## Notes

Non-breaking, and targeted at `2.0` only because that's where work is happening. Cherry-picks cleanly to `main` if a `1.21.1` is ever cut — your call.

Related to #539

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- `Gacela::resetCache()` now clears the memoized `class_exists()` answers and the reflection pool. `ClassValidator` caches the *negative* answer too, so a class that was not loadable when first resolved stayed "missing" for the life of the process — `Gacela::resetCache()` did not clear it, because `ClassValidator::resetCache()` and `ReflectionClassPool::reset()` were reachable only from their own unit tests. Affects any process where the set of loadable classes changes after the first resolution: long-running workers (RoadRunner, Swoole, queue consumers) that re-bootstrap, code generation, and `cache:warm` emitting classes
+- `Gacela::resetCache()` now drops the memoized "this class does not exist" answers. `ClassValidator` caches the *negative* result of `class_exists()`, so a class that was not loadable when first resolved stayed "missing" for the life of the process — `Gacela::resetCache()` did not clear it, because `ClassValidator::resetCache()` was reachable only from its own unit test. Affects any process where the set of loadable classes changes after the first resolution: long-running workers (RoadRunner, Swoole, queue consumers) that re-bootstrap, code generation, and `cache:warm` emitting classes. Positive answers are kept — a class that exists cannot stop existing, so they never go stale, and clearing them cost ~20% on the no-file-cache bootstrap path for no correctness gain
 
 ### Changed (BREAKING — 2.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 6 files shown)


Chemaclass:
## Revised after CI — both failures were the same finding

CI caught two things, and they turned out to be one better design.

**Perf guard: `FileCacheBench::bench_without_cache` +20.07%.** Not noise, and not a flaky gate — that was my fix. Clearing the whole memo means every cold resolution re-runs the autoloader for classes already known to exist. The other 21 gate subjects stayed inside ±6.5%, so the guard isolated it cleanly.

**Mutation testing: exactly one escaped mutant** — `MethodCallRemoval` on `ReflectionClassPool::reset()` in `Gacela::resetCache()`. It escaped because removing the call has no observable effect.

Both point the same way:

- **`class_exists()` returning `true` is permanent.** A class cannot stop existing within a process, so a positive answer never goes stale. Only the negatives can become wrong. `resetCache()` now drops just those — same bug fixed, cost gone.
- **`ReflectionClassPool` does not need resetting at all.** Reflection of a loaded class cannot change, and the pool is bounded by the set of loaded classes, so nothing goes stale and nothing leaks. It is now exempted with that reason rather than called. The honest way to kill that mutant was to stop making a call that does nothing — not to write a test that reaches into a private static to watch it happen.

### On what I did *not* test

There is deliberately no test asserting that positive answers survive. Once a class is defined, `class_exists()` never consults the autoloader again, so such a test could not distinguish a kept entry from a cleared one — it would pass either way and assert nothing. I wrote it, saw it was vacuous, and deleted it.

That property is observable only as cost, and `FileCacheBench::bench_without_cache` is what measures it. The perf gate is the regression test for the optimization; the reasoning is recorded in the test docblock so the next person doesn't add the vacuous version.

Verified locally: 100% MSI on both changed files (after clearing a bloated PHPStan result cache that was crashing infection's initial suite at the 128M subprocess limit), 1310 tests green, quality clean.

PR #557, opened 2026-07-27
fix(cache): clear the file caches' in-memory copy on resetCache

Chemaclass:
## 📚 Description

Step 2 of #539 — *"a guard test asserts every statically-stateful class is reset centrally"* — plus the leak it found on its first full-suite run.

`ResetCacheCoverageTest` already guards `Gacela::resetCache()`, but it starts from the resets that **exist** and asks who calls them. By construction it cannot see a class that holds static state and declares no reset at all, and #539 measured seven of those. It also discovers by method name, so a clear called `clearStaticCache()` is invisible to it.

`StaticStateCoverageTest` starts from the **state** instead: it enumerates every `static` property under `src/`, populates them by resolving a real module, calls `Gacela::resetCache()`, and requires each one to be back at its declared default — or to be listed with the reason its lifetime is not cache lifetime.

Written red first. It reported a real leak immediately.

### The leak

With `gacela-cache-enabled`, entries that `ClassNamePhpCache` and friends read from disk kept answering **after** an explicit `Gacela::resetCache()`. `GacelaConfig::resetInMemoryCache()` did not reset that layer at all.

`AbstractPhpFileCache::clearStaticCache()` has existed since the batching work and is unit-tested — but it clears one subclass at a time, and the central reset cannot name concrete subclasses. The only callers were the test suite and a benchmark. Exactly the "a working reset that nothing calls" shape as the `ClassValidator` and `ReflectionClassPool` cases before it, and the seventeenth cache bug of the family #539 tracks.

It only surfaces under the full suite, where an earlier test enables the file cache — so it is pinned separately by `AbstractPhpFileCacheResetTest`, which fails 4 of 5 without the wiring.

## 🔖 Changes

- `AbstractPhpFileCache::resetCache()` — keyless, so it clears every subclass without knowing any of them. Wired into `Gacela::resetCache()`
- `put()` recomputes a missing filename from the instance's own cache directory instead of trusting the registry, so the reset can drop the registry too. The old comment's reason for keeping it back no longer applies
- `all()` reads a missing slot as empty, so an instance that outlived a reset does not fault
- `ClassInfo::$callerClassCache` given its `= []` default — it was the one typed static with no default, and so could never be observed at rest
- `StaticStateCoverageTest` + its own fixture module (pillar resolution, an attribute-declared `#[Provides]` provider, `#[ServiceMap]` docblock service resolution and a `#[Cacheable]` method). The fixture belongs to the test rather than being borrowed from a feature suite, so nobody can defang the workload by editing a module whose own tests still pass
- `AbstractPhpFileCacheResetTest` — 5 regression tests on the leak above

### The declared exceptions

18 statics legitimately outlive a cache reset, each carrying its reason in `OUTLIVES_RESET`, in three kinds:

- **configuration** — `Gacela::$appRootDir`, `CacheableConfig::$storage`, `HealthCheckRegistry::$checks`. Clearing these would drop what the app asked for
- **pure memoization** — `GlobalKey::$cache`, `ReflectionClassPool::$cache`, `ProvidesScanner::$cache`. A total function of a key that cannot change within a process
- **unclearable by construction** — the `ServiceResolverAwareTrait` statics. A trait static exists once per using class, including application classes the framework never sees, so no central reset could reach them all. Safe only because they are pure memoization

The list is self-invalidating in both directions: an entry naming a property that no longer exists fails, and so does one the reset now clears after all.

### Verification

Each guard was made to fail before being trusted:

| removed | result |
|---|---|
| `AbstractFacade::resetCache()` from the central list | ✅ flags `AbstractFacade::$factories` |
| `AbstractFactory::resetCache()` | ✅ flags both its statics |
| `InMemoryCache::resetCache()` | ✅ flags `InMemoryCache::$cache` |
| a bogus `OUTLIVES_RESET` key | ✅ "no longer a static property" |
| an entry the reset does clear | ✅ "resetCache() now clears them" |
| `AbstractPhpFileCache::resetCache()` wiring | ✅ 4 of 5 regression tests fail |
| the filename recompute | ✅ errors on write-after-reset |

The workload populates 24 statics; the guard fails if it ever drops below 10, so it cannot degrade into asserting nothing.

Full suite green on seeds `1`, `555`, `31337`, `1111`, `4242`, `9999`, `7` — this test mutates global state, so order-independence was checked rather than assumed.

Related to #539

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,13 @@
 
 ### Fixed
 
+- **The in-memory copy of the file-backed caches survived `Gacela::resetCache()`.** With `gacela-cache-enabled`, entries that `ClassNamePhpCache` and friends had read from disk kept answering after an explicit reset, so `GacelaConfig::resetInMemoryCache()` did not reset that layer. `AbstractPhpFileCache::clearStaticCache()` had existed since the batching work and was unit-tested, but it clears one subclass at a time and the central reset cannot name concrete subclasses, so the only callers were the tests themselves. `AbstractPhpFileCache::resetCache()` now clears every subclass at once and `Gacela::resetCache()` calls it. The seventeenth of the cache bugs tracked in #539, and the first one found by a guard rather than by a user
+
 - **`extendService()` on an id that names an autowirable class threw instead of scheduling the extension.** A regression in `gacela-project/container` 1.0, shipped in Gacela by the 1.0 bump: `has()` moved to the PSR-11 question ("will `get()` resolve this?"), which is `true` for any instantiable class, so `extend()` took the already-defined branch. Fixed by requiring container `^1.1`. Gacela's own tests missed it because every existing `extendService()` case uses a plain string id, and the bug only bites when the id happens to name a real class
 
 ### Internal
 
+- `StaticStateCoverageTest` guards process-global state from the opposite end to `ResetCacheCoverageTest`: it enumerates every `static` property under `src/`, populates them by resolving a real module, calls `Gacela::resetCache()` and requires each one to be back at its declared default or listed with the reason its lifetime is not cache lifetime. The existing guard starts from the resets that exist, so by construction it cannot see a class holding state with no reset at all — seven of those are what #539 measured, and an eighth (above) was found by writing this one
 - Raised the `nikic/php-parser` floor to `^5.4`. Psalm 6.16 declares `^5.0.0` but reads `Property::$hooks`, which only exists from 5.4, so a `--prefer-lowest` install produced a Psalm that crashes on any file containing a property. Nothing had noticed because nothing ran Psalm from inside the test suite until the plugin test did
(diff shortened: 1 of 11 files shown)


Chemaclass:
Follow-up in `8e34...`: the escaped `AssignCoalesce` mutant on `??=` was pointing at a real hole rather than asking for a test.

`commitBatch()` is static, so it has no instance to ask for a filename and resolves each dirty class through `self::$filenames`, skipping any it cannot find. Since `resetCache()` now empties that registry, a batch opened **after** a reset would mark the cache dirty and then be silently dropped on commit — writes lost, no error anywhere.

`put()` now re-registers the filename unconditionally, which closes the hole and removes the memo the mutant was aimed at. Pinned by `test_a_batch_opened_after_a_reset_is_still_flushed_to_disk`, verified to fail without the fix.

Issue #503, opened 2026-07-24
epic: Gacela 2.0 roadmap — attribute-first + one container


Chemaclass:
## Gacela 2.0 — roadmap

Tracking issue for the 2.0 effort. **Direction: turn on the modern DI engine already shipped, and make module wiring explicit and type-safe (attribute-first).**

**Work happens on `main`**, which is the 2.0 line and is cut as 2.0.0 when this list is done. There is no `2.0` branch and no `1.x` maintenance branch — 1.21.0 is the final 1.x release, and its tag is what consumers resolve against.

### Principle

The container (`gacela-project/container`) is already a full autowiring/attribute engine — `#[Inject]/#[Singleton]/#[Factory]/#[Lazy]`, contextual bindings, tags, a compiler. The framework used to wire it as a stringly-typed service locator behind runtime naming-magic. 2.0 makes the wiring explicit and routes construction **through** the container.

### Status: 2.0 is one command away

Everything below the "Done" heading is merged on `main`, all four workflows (Tests, Code style, Infection, Module cycles) are green at `c408893b`, and the release is `./release.sh 2.0.0` — it rewrites `## Unreleased` to the version header, tags, and cuts the GitHub release, refusing to run unless the tree is clean, on `main`, in sync with origin, and CI is green for HEAD.

- [ ] **Cut 2.0.0** — `./release.sh 2.0.0`

Nothing else is open against this milestone. `UPGRADE.md` is written and its inaccuracies were corrected in #563 and #588, `README` states the 8.3 floor and links the migration, `docs/upgrading.md` is a pointer so old links survive, and 1.21 is documented as the final 1.x in `8647a0c`.

### Shipped in 1.20 — non-breaking groundwork

- [x] #499 — route domain construction through the module container (`AbstractFactory::make`) — **keystone autowire-first lever** (PR #505)
- [x] #501 — `make:module --minimal` + optional Config/Provider pillars (PR #506)
- [x] #502 — scaffolder provider template uses `#[Provides]` (PR #504)

### Shipped in 1.21 — static analysis and console guards

Landed in 1.x on purpose, so they are available **before** you migrate rather than as a reward for migrating.

- [x] #521 — PHPStan types `getFacade()/getFactory()/getConfig()` from `#[ServiceMap]` instead of shipping suppressions
- [x] #522 — the module graph is enforcing: cycle detection, edge classification, exit code
[...]

PR #558, opened 2026-07-27
feat(container): adopt container 1.2 typed stats and property injection

Chemaclass:
## 📚 Description

Bumps `gacela-project/container` to `^1.2` and adopts the two additions worth taking rather than merely tolerating. No breaking changes this time — the upgrade itself is a one-line constraint change, and everything below is deliberate adoption.

### Typed stats, and why it matters here

`debug:container` indexed six string keys of the array returned by `getStats()`. Upstream carves that array's shape out of its backward compatibility promise — *"Do not build logic on it"* — and Gacela did. `ContainerStatsShapeTest` existed only to move the resulting breakage onto our CI instead of into a user's terminal as an undefined-array-key error.

Container 1.2 removes the need for that trade. `stats()` returns a `final readonly ContainerStats` whose properties **are** covered, so the command reads typed properties and a rename now fails at analysis time. Memory comes back as `memoryUsageBytes` (an int, comparable and summable) with `memoryUsageFormatted()` for display.

`getStats()` is untouched and keeps working for the whole of 1.x.

`Container::stats()` is forwarded explicitly, because `stats()` lives on the concrete container and **not** on `ContainerInterface` — so unlike every other method in that class, it is not held in place by the interface, which is the mechanism the forwarding relies on everywhere else. `ContainerStatsShapeTest` is rewritten rather than deleted to pin exactly that: the array's keys are no longer the useful claim, the delegation still is.

### `#[Inject]` on properties

Reaches Gacela through `AbstractFactory::make()` and container resolution with no framework change, the same way `#[Lazy]` did. It exists for classes whose constructor is not yours to change — a base class from a vendor package, or one whose signature is fixed by a framework contract.

Constructor injection stays the default everywhere else, and the docs say so: a dependency in the signature is visible to a reader, and to a plain `new` outside the container.

## 🔖 Changes

- `composer.json`: `gacela-project/container` `^1.1` → `^1.2.1` (see *A regression the gate caught* below for why the floor is the patch, not `^1.2`)
- `Container::stats()` forwarded; the renamed `ContainerStats` → `StatsArray` psalm alias fixed on the import
- `ConsoleFactory` / `ConsoleFacade` return `ContainerStats` instead of an array, and the hand-written `@psalm-type ContainerStats` array alias is deleted — the type now comes from the package
- `DebugContainerCommand` reads typed properties and `memoryUsageFormatted()`
- `ContainerStatsShapeTest` rewritten to pin the delegation rather than the array shape
- `InjectOnPropertiesTest` — property injection, inherited private properties, and the `readonly` rejection asserted on its actual message rather than loosely
- `docs/container-configuration.md` — a new *On properties* section covering the three rejected cases and the cycle note

### Also free in the bump

Upstream reports resolution 5–8% faster on the chain benchmarks (the class plan is passed into instantiation rather than re-read per node), and classes without `#[Inject]` properties never enter the injection path.

### A regression the gate caught

The first push failed **Performance regression guard** on all four `ContainerResolutionBench` subjects — `bench_resolve_no_dependencies` at +17.59%, the others between +10.29% and +10.75%, with peak memory up 3.5–4%. Not noise, and not any of the code in this PR: reproduced locally, and bisected to the instantiability guard container added in **1.1.1**.

That guard builds a `ReflectionClass` of its own to decide whether a class can be instantiated. `DependencyResolver::describeClass()` reflects the same class a moment later and already stores the answer in its plan, so it was a second reflection for a question that was about to be free — and its cache was per container, so a cold `Container::get()` paid it every time. `ContainerResolutionBench` constructs a fresh container per revolution deliberately, to measure exactly that path, which is why it took the full hit.

Nothing here had noticed because this repository's lockfile still pinned 1.1.0. Raising the floor is what exposed a cost that has been shipping since 1.1.1.

Fixed upstream in [container#105](https://github.com/gacela-project/container/pull/105), released as 1.2.1: the resolver answers `isInstantiable()` off the plan it already produces, and the verdict is memoized process-wide — positives only, since a loaded class definition cannot change while a `class_exists()` miss can start succeeding later.

Verified on a CI runner before 1.2.1 was cut, against the same `main` baseline:

| subject | before | after |
|---|---|---|
| `bench_resolve_no_dependencies` | +17.59% | **+6.79%** |
| `bench_resolve_with_inject_attribute` | +10.29% | **+7.97%** |
| `bench_resolve_with_bindings` | +10.75% | **+6.05%** |
| `bench_resolve_deep_chain` | +10.41% | **+4.59%** |
| peak memory | +3.5–4% | **+0.6–0.7%** |

Related to #503


Chemaclass:
⚠️ **The performance gate is red, and it looks real rather than noisy.** Not merging on that.

`ContainerResolutionBench` regressed uniformly against the 1.1 baseline:

| bench | time |
|---|---|
| `bench_resolve_no_dependencies` | **+17.59%** |
| `bench_resolve_with_bindings` | +10.75% |
| `bench_resolve_deep_chain` | +10.41% |
| `bench_resolve_with_inject_attribute` | +10.29% |

The gate trips at >10%. What makes this worth investigating rather than re-running: the other benches on the same job show wild variance swings (`±` moving +112%, +132%) that do read as runner noise, but the four container benches moved **coherently in one direction**, which noise does not usually do. Memory is also up ~3.5-4% across the same four and flat everywhere else.

It is also the opposite of the direction 1.2 advertises — upstream reports resolution 5-8% *faster* on the chain benchmarks, from passing the class plan into instantiation instead of re-reading it per node. So either the benchmark is measuring something upstream's is not, or the property-injection scan is costing more on first touch than the memo saves back.

Worth ruling out before merging:

1. A local A/B of `ContainerResolutionBench` on 1.1 vs 1.2, off CI hardware, to confirm the regression exists at all
2. Whether the cost is per-class-first-touch (the new property scan) rather than per-resolution — the benches use `revs=200`, so a one-off cost should amortize and this should not show
3. Whether Gacela's `Container` composition wrapper interacts badly with the new instantiation path

Everything else on the PR is green: 23 checks pass, full suite green, quality clean.

PR #559, opened 2026-07-27
test: validate container instantiability-guard fix against the perf gate

Chemaclass:
Throwaway. Pins `gacela-project/container` to the branch in gacela-project/container#105 so the **Performance regression guard** runs on a CI runner with a fresh `main` baseline.

Closed and deleted once the gate result is read. Not for merge.


Chemaclass:
Validated: the perf gate passes with the fix from gacela-project/container#105 (worst subject +17.59% → +6.79%, all four under the ±10% gate). Throwaway branch, closing.

Issue #163, opened 2022-05-30
Adding cache for DocBlockResolverAwareTrait


Chemaclass:
There is only **one place** where Gacela is using reflection: `DocBlockResolverAwareTrait`, and this is a very special trait that could be used in your infrastructure layer to get the facades of your modules, for example.

The thing is that nothing stops the user to use this trait whenever they want. Anyway, the goal is to build a file cache mechanism to avoid using reflection on "production" code.

> This issue is to give it another try this PR ➡️  https://github.com/gacela-project/gacela/pull/120 (We can use this idea to get some inspiration).

Chemaclass:
Actually... not yet. We created a cache system for the class resolver itself https://github.com/gacela-project/gacela/pull/165 but we are still using reflection for the `DocBlockResolverAwareTrait`.
Now we need to think how could we create something similar (I don't know how yet) to have a cache as well for this trait in order to avoid using reflection every time... 🤔 I will think about this this week.

Chemaclass:
Done: https://github.com/gacela-project/gacela/pull/168

PR #586, opened 2026-08-01
chore(deps): update to container 2.0 and drop the closure-wrapping layer

Chemaclass:
## 📚 Description

`gacela-project/container` `^1.5.0` → `^2.0.0`. Everything else is already current.

2.0 puts the **whole surface on `ContainerInterface`**, which is the change that matters here: an unforwarded method is now a compile error instead of something that silently goes unreachable. Two capabilities that were unreachable for the whole of 1.x shipped that way.

Four breaking changes needed handling, and two of 2.0's additions **replace code here outright**.

## ✂️ `withSelfReference()` removes the closure-wrapping layer

The inner container passed *itself* to service closures, so Gacela re-wrapped every user closure to substitute the decorator — otherwise the documented provider signature `static fn (Container $c) => $c->getLocator()` would fatal, because `getLocator()` exists only on the decorator.

That wrapping then broke a second invariant: `factory()` and `protect()` mark closures **by identity**, so a re-wrapped closure silently lost its mark. Hence a `WeakMap` tracking every wrapper, threaded through eight call sites — ~50 lines, 29 touchpoints.

All of it is replaced by telling the container which object to hand out:

```php
$this->inner = (new GacelaContainer(...))->withSelfReference($this);
```

The `WeakMap` goes with it, and so does the leak it was fixed for in #582.

Verified end to end, since this is the load-bearing part:

```
set() closure  -> decorator reached getLocator()
factory fresh  -> true      (identity mark survives without our wrapping)
protect same   -> true
```

## 🌳 `createScope()` is forwarded

It was the decorator's one recorded non-forward for all of 1.x, and the reason was exactly this: forwarding the raw child would have handed *its* closures the inner container. With the seam, a scope is decorated like its parent and keeps the Locator and lifecycle events.

Gacela still does not create scopes itself — which lifetime owns one is an open design question (#539) — but the primitive is reachable, and `ContainerForwardingCoverageTest::NOT_FORWARDED` is now **empty**. That test's own staleness check, added in #585, is what flagged the entry the moment it stopped being true.

## 🔖 The rest

- `load()`/`loadFile()` return the ids they registered and take an optional per-id callback
- `ContainerStats::memoryUsageFormatted()` → `processMemoryFormatted()`; `debug:container` now labels it **Process Memory**, which is what it always measured

## ⏸️ Deliberately not taken

2.0's `afterResolving()` matches instances the same way Gacela's does, so delegating looked like a free ~40-line deletion. **It is not behaviour-preserving:** upstream fires for *any* resolved value, where this one guards on `is_object()`. Two tests caught it immediately — a typed callback like `fn (LoggerAwareInterface $s) => …` receives a `string` and TypeErrors, and a `null` resolution reached a hook.

Reconciling the two is a behaviour decision, not part of a dependency bump. The local implementation stays, with the reason recorded on the property so the next reader does not retry it.

## ✅ Checks

`composer test` green (940 unit / 258 integration / 296 feature). New `ContainerScopeTest` covers the scope being decorated, resolving through its parent, isolation between siblings, and shadowing without mutating. `ClosureWrapperReleaseTest` is deleted — the machinery it guarded no longer exists.

Also held, both re-verified rather than assumed:

- **`phpunit` stays on `^10.5`** — the recorded reason still holds.
- **`psalm/plugin-phpunit` stays on `^0.19`** — `0.20` needs `psalm-plugin-api ^0.1`, which conflicts with `vimeo/psalm <7.0.0`, and Psalm 7 is still at `beta19`.


Chemaclass:
## ⛔ Holding this: container 2.0 has a resolution regression

The performance gate fails on all four `ContainerResolutionBench` subjects, and it is **not** this PR's doing.

Reproduced locally against a `main` baseline on the same machine:

| subject | 1.5.0 | 2.0.0 | |
|---|---|---|---|
| `bench_resolve_no_dependencies` | 1.527μs | 1.839μs | +20.4% |
| `bench_resolve_with_inject_attribute` | 2.973μs | 3.428μs | +15.3% |
| `bench_resolve_with_bindings` | 2.964μs | 3.759μs | +26.8% |
| `bench_resolve_deep_chain` | 4.953μs | 5.533μs | +11.7% |

CI measured +18.6 / +16.5 / +23.8 / +11.1% on its own runner, so it is not local noise.

**Isolated from `withSelfReference()`**, the obvious suspect: disabling that call and re-running against the same baseline gives an identical result (+20.6 / +14.5 / +28.0 / +11.5%). The seam costs nothing here. Everything else in this branch *removes* work.

Filed as [container#181](https://github.com/gacela-project/container/issues/181) with the isolation, alongside the existing [#163](https://github.com/gacela-project/container/issues/163) which already tracks ~7% of drift in 1.5.0.

The code side of this PR is finished and green — quality, 940/258/296 tests, module cycles, type checker. Only the perf gate is red, and re-baselining it would bank a 11-28% regression as the new normal.

Suggest holding until #181 lands, then re-running. If you would rather take the API wins now and re-baseline, say so and I will.

PR #590, opened 2026-08-01
ref: module containers as scopes of one app container (closes #539)

### LINKED ISSUE #539: ref: one container with an explicit scope seam (2.1)

Chemaclass:
## 📚 Description

Step 3 of #539. `AbstractFactory` built one container per Factory class through `Container::withConfig()`, which walks the whole of `gacela.php` — every binding, factory, alias, contextual binding, tag and hook — and did it again for each. 79 walks in this repository alone, and the cost grows with an application's wiring rather than with its number of modules.

Each module now gets a **scope** of the container bootstrap already built.

## 🔖 What it took

- **Scopes, not copies.** Registration is not copied and a miss falls through, so a module still cannot see a sibling's provider keys — which is what stops two providers using the same un-namespaced key from colliding. Module isolation is unchanged: two modules resolving the same app-wide binding still get one instance each.

- **`extendService()` needed handling** — the "parent-owned vs scope-owned" question the issue flagged. It is app-wide configuration, but the service it decorates is often registered by a *module's* Provider into that module's scope, where a parent-held extension can never reach it. Each scope schedules those itself, skipping ids the parent owns; asking a scope to extend an inherited instance is refused upstream rather than silently ignored.

- **One container, not two.** The first pass shared a parent between Factory classes but built it itself, so `gacela.php` was still walked twice — once by bootstrap, once here. `SharedAppContainerTest` counts `BindingRegisteredEvent` across two modules and caught it: the app-wide binding was announced twice. `Gacela::bootstrap()` now hands its container down.

- **Pushed, not pulled.** Reaching for `Gacela::container()` from a Factory makes the framework's entry point a dependency of every module factory. `NoCircularDependenciesTest` rejected that as a cycle, correctly, so the container is handed down instead.

## 📊 The tradeoff

The saving is on container construction and scales with how much configuration there is to re-walk. 79 containers:

| app-wide entries | per-container | scopes |
|---|---|---|
| 25 | 1.553ms | 0.063ms |
| 300 | 18.04ms | 0.069ms |

The scope column is flat. This repository has almost no app-wide configuration, so it sees none of that and pays ~2.5% on warm class resolution for the fall-through — measured against `main` on container 2.0.1:

| benchmark | main | scopes | |
|---|---|---|---|
| `bench_class_resolving` | 1.101μs | 1.128μs | +2.5% |
| `bench_bootstrap_cold_rescan` | 42.43μs | 42.33μs | −0.2% |
| `bench_bootstrap_warm` | 27.62μs | 27.58μs | −0.2% |

So this pays for an application with real wiring and is roughly free for one without.

An earlier revision of this PR measured +5-7% and was held as a draft; that was against container 2.0.0, before [container#181](https://github.com/gacela-project/container/issues/181) was fixed and adopted in #592.

## ✅ Tests

`SharedAppContainerTest` pins that the app configuration is registered **once** across modules, and that sharing the parent does not cost a module its access to the binding. Full gate green: quality, 946 / 265 / 296, module cycles, mutation, perf.

Closes #539

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,29 @@ Migration is three mechanical renames. See [UPGRADE.md](UPGRADE.md), and run
 
 ### Changed
 
+- **Module containers are scopes of one app container.** `AbstractFactory` built
+  one container per Factory class through `Container::withConfig()`, which walks
+  the whole of `gacela.php` — every binding, factory, alias, contextual binding,
+  tag and hook — and did it again for each. 79 walks in this repository alone,
+  and the cost grows with an application's wiring rather than with its number of
+  modules. Each module now gets a **scope** of one shared app container,
+  carrying only its own Provider.
+
+  Registration is not copied and a miss falls through, so a module still cannot
+  see a sibling's provider keys — which is what keeps two providers using the
+  same un-namespaced key from colliding. Module isolation is unchanged: two
+  modules resolving the same app-wide binding still get one instance each.
+
+  `extendService()` is app-wide configuration but often decorates a service a
+  *module's* Provider registers, into that module's scope where a parent-held
+  extension cannot reach it. Each scope now schedules those itself, skipping
+  ids the parent owns.
+
+  The saving is on container construction and scales with how much there is to
+  re-walk: 79 containers against 25 app-wide entries go 1.55ms → 0.06ms, and
+  against 300 entries 18.0ms → 0.07ms — the scope column is flat. An application
+  with little configuration sees none of that and pays ~2.5% on warm class
+  resolution for the fall-through.
 - **`gacela-project/container` `^2.0.1`** (was `^0.10.0`). `Container` is
   `final`, so Gacela decorates it by composition. Two of its fixes land in
   Gacela's own path: a class-string sharing a name with a function was *invoked*
--- a/src/Framework/AbstractFactory.php
+++ b/src/Framework/AbstractFactory.php
@@ -25,6 +25,23 @@ abstract class AbstractFactory
     private static array $containers = [];
 
     /**
+     * The app-wide configuration, resolved once and shared by every module.
+     *
+     * Each module container used to be built by `Container::withConfig()`, which
+     * walks the whole of `gacela.php` -- every binding, factory, alias,
+     * contextual binding, tag and hook -- and does it again per Factory class.
+     * Only the module's own Provider differs. That is 79 walks in this
+     * repository alone, and the cost grows with the application's wiring rather
+     * than with the number of modules.
+     *
+     * A module now gets a *scope* of this instead: registration is not copied,
+     * a miss falls through, and anything the module registers shadows the parent
+     * for that module alone -- which is what keeps two providers using the same
+     * un-namespaced key from colliding.
+     */
+    private static ?Container $appContainer = null;
+
+    /**
      * Modules that resolved no Provider at all. Their container is still usable
      * for make(), which autowires by type, but getProvidedDependency() has
      * nothing to read from and reports the missing Provider instead.
@@ -43,6 +60,7 @@ abstract class AbstractFactory
     {
         self::$containers = [];
         self::$providerless = [];
+        self::$appContainer = null;
     }
 
     /**
@@ -105,7 +123,8 @@ abstract class AbstractFactory
 
     private function createContainerWithProvidedDependencies(): Container
     {
-        $container = Container::withConfig(Config::getInstance());
+        $container = self::appContainer()->createScope();
+        self::scheduleAppWideExtensions($container);
 
         $resolver = (new ProviderResolver())->resolve($this);
         $resolver?->register($container);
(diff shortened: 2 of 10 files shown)


Chemaclass:
## Re-measured on container 2.0.1 — taking it out of draft

The numbers above were taken against container 2.0.0, whose arg-builder regression ([container#181](https://github.com/gacela-project/container/issues/181)) has since been fixed and adopted in #592. That moved the floor this change is measured on, so the draft's conclusion was stale.

Rebased on current `main` and re-run:

| benchmark | main | scopes | |
|---|---|---|---|
| `bench_class_resolving` | 1.101μs | 1.128μs | **+2.5%** |
| `bench_bootstrap_cold_rescan` | 42.43μs | 42.33μs | −0.2% |
| `bench_bootstrap_warm` | 27.62μs | 27.58μs | −0.2% |
| `FileCacheBench::without_cache` | 365.0μs | 379.3μs | +3.9% |
| `FileCacheBench::with_cache` | 250.7μs | 251.4μs | +0.3% |

Worst case **+3.9%**, comfortably inside the ±10% gate, where on 2.0.0 it was +5-7% and marginal.

## The tradeoff, stated plainly

This repository still shows **no gain**, because it has almost no app-wide configuration — and the saving is entirely on re-walking that configuration. Measured against a growing config, 79 containers:

| app-wide entries | per-container | scopes |
|---|---|---|
| 25 | 1.553ms | 0.063ms |
| 300 | 18.04ms | 0.069ms |

The scope column is flat. So this pays for an application with real wiring and costs ~2.5% on warm class resolution for one without.

Full gate green: quality, 946 / 263 / 296, module cycles.

Related to #539

PR #729, opened 2026-08-13
✨ Refuse to overwrite an existing module

Chemaclass:
## 📚 Description

`make:module App/Wallet` on a module that already exists replaced all four
pillars with stubs and reported this:

```
> Path 'src/Wallet/WalletFacade.php' created successfully
> Path 'src/Wallet/WalletFactory.php' created successfully
> Path 'src/Wallet/WalletConfig.php' created successfully
> Path 'src/Wallet/WalletProvider.php' created successfully
Module 'Wallet' created successfully
```

No prompt, no flag, no mention that anything had been there. "created" is not
even what happened. `make:file` did the same. The only record that hand-written
code had been there was the file that had just been written over.

`MakeModuleCommand` already states the right rule, for the unusable-path check
directly above this one:

> *a half-written module is worse than a refusal*

Now:

```
src/Wallet/WalletFacade.php, src/Wallet/WalletFactory.php, ... already exist.
Nothing was written. Pass --force to replace them.
```

## 🔖 Changes

- Every target is checked before the first one is written, so a run that would
  replace something writes nothing. `--force` (`-f`) opts in.
- The refusal names the files in the way — including the partial case, where a
  module missing one pillar reports the three that are there — so `--force` is an
  informed choice rather than a retry.
- `FileContentGenerator::targetPath()` names where `generate()` would write
  without writing it, and without creating the directory on the way (pinned by a
  test: a refused run must not leave an empty module directory behind).
- The three template branches in `MakeModuleCommand` collapse into one list of
  `[filename, subDirectory]` pairs, read once to check and again to write.
- Singular/plural wording lives in `ConsoleSection` so both makers refuse in the
  same words.

## ⚠️ Behaviour change

A script that regenerates a module in place now needs `--force`. Called out in the
changelog under `### Changed`.

## ✅ Notes

**I wrote a vacuous test and caught it by breaking the code.** My first invariant
test asserted that `targetPath()` agrees with where `generate()` writes — but
`generate()` builds its path *by calling* `targetPath()`, so they cannot disagree
and the assertion was a tautology. It passed with `targetPath()` deliberately
broken. Agreement is a property of the design, not something a test can add;
what is worth pinning is that the builder is right, so it now asserts the literal
path for each shape (short name, sub-directory, both). Verified it fails on the
same deliberate break: 2 of 4.

- Gacela's own `GacelaFacadeOnlyDelegates` rule caught my first version putting
  generator selection inline in the Facade. Moved to the Factory.
- Verified by hand end-to-end: refuse (exit 1, file byte-identical), `--force`
  (exit 0, replaced), fresh module unaffected, service template with `--with-tests`
  unaffected, partial module lists only what blocks it.
- Full gate green: 1700 unit / 335 integration / 418 feature, psalm + phpstan +
  cs-fixer + rector clean. Infection on the changed domain and IO files: 32
  mutants, **0 escaped, 100% MSI** (the `Command` classes are globally ignored as
  presentation code).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 ### Changed
 
+- Refuse to overwrite in `make:module` and `make:file`, and add `--force` for when replacing is the intent. Generating over an existing module replaced every pillar with a stub and reported "created successfully" for each one, with no prompt and no flag — the only record that hand-written code had been there was the file it had just been written over. Every target is now checked before the first is written, so a run that would replace something writes nothing and names what is in the way. **This changes existing behaviour**: a script that regenerates a module in place now needs `--force`
 - Resolve a class name to its kind through the configured suffixes rather than the four literal pillar names. A project on `addSuffixTypeFacade('PublicApi')` now normalizes `App\Foo\FooPublicApi` to the `Facade` key the resolver looks up, so an `overrideExistingResolvedClass()` call on such a name takes effect where it was inert before
 
 ### Fixed
--- a/src/Console/ConsoleFacade.php
+++ b/src/Console/ConsoleFacade.php
@@ -52,19 +52,22 @@ final class ConsoleFacade extends AbstractFacade
     /**
      * Which of the files a `make:*` run would write already exist.
      *
+     * A target path is built from the module arguments alone -- the stubs a
+     * template would fill in never reach it -- so which template is in play
+     * makes no difference here, and asking would be a distinction nothing
+     * could observe.
+     *
      * @param list<array{string, string}> $files [filename, subDirectory] pairs
-     * @param bool $service whether the service template's generator is the one that would write them
      *
      * @return list<string>
      */
     public function existingGeneratedFiles(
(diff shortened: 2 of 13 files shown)


Chemaclass:
The mutation gate caught real over-engineering rather than a missing test.

Three mutants escaped, and all three were unkillable for the same reason: `existingGeneratedFiles()` took a `$service` flag to pick between the two generators, but a target path is built from the module arguments alone — the stubs a template would fill in never reach it. Both generators produce identical paths, so swapping the branches of `createFileContentGeneratorFor()` changed nothing any test could observe. There was no behaviour there to assert.

So I deleted the branch instead of testing it: the flag, the selector method, and the parameter defaults no caller relied on are gone. A feature test now covers the service template's `Domain/` and `Tests/` sub-directories resolving through the plain generator, which is the agreement this actually rests on.

16 mutations → 10, all killed, 100% MSI. Full gate green: 1700 unit / 335 integration / 419 feature.

PR #730, opened 2026-08-13
🐛 Never touch the filesystem root when there is no cache directory

Chemaclass:
## 📚 Description

`FileCache::clear()` and `stats()` built their glob as `$this->directory . '/*.php'`.
`normalizeDirectory()` reduces `''`, `'/'`, `'//'` and whitespace all to an empty
string, so that pattern becomes `'/*.php'` — every PHP file at the filesystem root.
`stats()` reported them as cache entries; `clear()` would have unlinked them.

Verified by construction (`clear()` never called):

```
empty string   -> directory=''  persistent=no
root           -> directory=''  persistent=no
whitespace     -> directory=''  persistent=no
double slash   -> directory=''  persistent=no
```

`persistent=no` is the part that makes this indefensible: the cache knows the
directory is unusable, writes nothing, and holds nothing on disk — and would still
delete. The class docblock states the contract it was breaking:

> *Degradation: an unusable directory disables persistence; entries then live only in memory.*

`FileCache` is a public primitive consumers instantiate directly, so a config value
that resolves to empty — an unset env var, a blank setting — reaches this.

## 🔖 Changes

- `CacheFilePattern::forDirectory()` returns `null` for an empty directory and the
  `*.php` pattern otherwise. `clear()` and `stats()` go through it.
- A directory that merely *does not exist* is deliberately not a special case: the
  pattern is well-formed and glob matches nothing. Only "nowhere" needs the rule.

## ✅ Notes

**Why this is a class and not an early return.** My first version was
`if ($this->directory === '') { return []; }` inside `FileCache`. The tests for it
passed *without the fix*, because this machine's `/` holds no `.php` files — and the
mutation gate said the same thing more precisely: `ReturnRemoval` deleted the guard
and nothing failed. I could not write a portable test that fails on the unfixed
code without creating files at `/`.

Rather than ship a guard nothing watches, I moved the decision to where it is
observable. As a pure function the pattern *is* the return value, so "nowhere, not
the root" is asserted directly, and removing the guard now fails a test. **12
mutants on the changed lines, 0 escaped, 100% MSI** (up from 1 escaping).

- I chose `=== ''` over `isPersistent()` deliberately: `loadEntry()` falls through
  to `readEntry()`, so a read-only directory's files are still readable. Gating
  `stats()` on persistence would have it deny files that `get()` returns.
- Full gate green: 1718 unit / 335 integration / 419 feature, psalm + phpstan +
  cs-fixer + rector clean. Rebased onto `main` post-#729.


Chemaclass:
**Windows CI caught a deeper version of this, and my first fix was incomplete.**

Three Windows jobs failed. One was a test-portability issue; the other was real:

```
test_clearing_a_cache_with_nowhere_to_write_still_empties_memory
Failed asserting that 'value' is null.
```

`writeEntryToDisk()` has no persistence guard — it writes unconditionally. With an empty directory the entry path is `'' . DIRECTORY_SEPARATOR . <sha1> . '.php'`, i.e. a path at the **drive root**, which is writable on the Windows runner. So entries really were persisted to `D:\`, and because my first fix guarded only the glob, `clear()` no longer removed them and the next `get()` read them straight back.

Guarding the glob alone would have made the leak permanent. The scope is now every disk operation — read, write, delete, and the batch commit's index-file lock — via `CacheFilePath`, which answers "nowhere" for all of them.

Also from the failures:
- `'//'` is **not** "nowhere" on Windows: it normalizes to the UNC prefix `\\`, a different case on a different platform. Removed from the portable providers and documented, rather than special-cased.
- The mutation gate then flagged the batch-commit ternary (`? false :`) as unkillable. The fallback is equivalent to doing nothing when there is no directory, so it is an early return now — clearer, and one fewer branch. Covered by a test that commits a batch with nowhere to write.

36 mutants on the changed lines, **0 escaped, 100% MSI**. Full gate green: 1719 unit / 335 integration / 419 feature.

PR #770, opened 2026-08-13
fix: hold APP_ENV to the same alphabet as a config dimension

Chemaclass:
## 📚 Description

A declared config dimension is validated. `APP_ENV` is not — and it reaches the
same two places.

```bash
APP_REGION=../../etc php app.php
# ConfigDimensionException: The config dimension "APP_REGION" has the value
# "../../etc", which cannot be used: a dimension reaches both a glob pattern
# and a cache filename, so it may contain only letters, digits, "_", "." and "-".

APP_ENV=../../etc php app.php
# accepted
```

`APP_ENV` is the first link of that same chain — it names
`config/app-{APP_ENV}.php` and is part of the merged-config cache filename. Two
things observed in a scratch project before this change:

- `APP_ENV=../escaped` — the merged-config cache was written into a directory
  the framework created inside the cache dir, literally named
  `gacela-merged-config-<hash>-..`
- `APP_ENV=x/../../pwned` — the write silently failed, so the application booted
  **uncached on every request** with nothing to say why. Cache writes are
  best-effort by design, which is right for a permissions problem and unhelpful
  for a malformed name

`eu*` and `a b` were accepted too, and go into a glob pattern.

The value is normally operator-set rather than user-supplied, so this is a
configuration-integrity problem rather than a remote one — but it is the same
hazard the framework already decided to guard for every *other* variable in the
chain, and the pattern's own docblock was written about this one:

> Wide enough that no plausible existing `APP_ENV` stops working, narrow enough
> that `/` and `..` cannot walk out of the config directory.

## 🔖 Changes

- `ConfigDimensions::assertValueCanReachAPath()` holds the rule; both
  `fromEnvironment()` and `AppEnv::current()` call it, so the two cannot drift
- Checked in `AppEnv::current()` rather than at bootstrap: nothing reads
  `APP_ENV` past that method, so the answer and its validity stay in one place
- Nine cases: five rejected (`../escaped`, `x/../../pwned`, `/etc/passwd`,
  `eu*`, `a b`) and four accepted (`prod`, `staging-eu`, `local_dev`, `1.2`) —
  the alphabet a dimension already allows, so an ordinary environment name is
  unaffected
- `getting-started.md` says `APP_ENV` is held to the same alphabet

Both files: **100% covered-code MSI**, no escaped mutants.


Chemaclass:
Perf guard flagged this on the first run (`bench_bootstrap_warm` +9.00%, `bench_bootstrap_cold_rescan` +7.91%). Measured rather than assumed, since this change genuinely does add work to a bootstrap path:

- `AppEnv::current()` is called **twice** per bootstrap, so the change is two `preg_match` calls on a short string. That cannot be 9% of a 28 µs bootstrap.
- Interleaved A/B locally (best-of-7 × 400 bootstraps, swapping only this file):

| round | baseline | patched |
|---|---|---|
| 1 | 65.65 µs | 67.15 µs |
| 2 | 66.02 µs | 66.55 µs |

Consistently positive but ~0.5–1.5 µs, i.e. **≈1–2%**, against a ~2.3% run-to-run noise floor on this harness. The CI number is noise; the real cost is at or below what this harness can resolve.

I deliberately did **not** optimise it further. Skipping the check when `APP_ENV` is empty would save the regex for projects that never set it, but the empty string passes the pattern anyway — so that guard is behaviour-equivalent and creates a mutant nothing can kill, which this repo prefers to avoid over an `ignore` entry. Memoising the validated value is worse: `AppEnvTest` changes `APP_ENV` per test, and re-bootstrapping under a different env is supported.

PR #858, opened 2026-08-16
fix(console): stop calling a skipped path scanned

Chemaclass:
## 📚 Description

`Scanned:` (added in #854) listed every `appModulePaths` entry, including the ones discovery **skipped** because they name something that is not a directory. That states what did not happen — to a reader looking at an empty module list precisely because it did not. The line was added to answer exactly that reader's question, so getting it wrong is worse than not having printed it.

Before:

```
No modules found.
Scanned: real, not-there
```

After:

```
No modules found.
Scanned: real
Not scanned, not a directory: not-there
```

## 🔖 Changes

- `scannedModulePaths()` returns only entries that were actually walked; new `unscannedModulePaths()` returns the rest
- They print on separate lines because they need opposite reactions: one says where to look for the missing module, the other that the configuration names somewhere that is not there
- Discovery already `trigger_error`s about these, but on a stream a command's own output may not share, and which a reader scanning a report has no reason to be watching. The test captures that warning and asserts it, since phpunit here is configured `failOnWarning="true"` and a stray one fails the run


Chemaclass:
Rebased onto `main` with #857 and extended: `doctor` grew its `Scanned:` line from that PR while `scannedModulePaths()` still returned every configured entry. With this PR narrowing that accessor, doctor would have dropped the skipped paths **silently** — the same failure as calling them scanned, one report further along, since every check runs without those paths and a screen of ticks is what makes it look fine.

Doctor now prints `Not scanned, not a directory:` under its scope line and carries an `unscanned` key beside `scanned` in `--format=json`.

Neither PR was wrong on its own; the gap only existed in the combination, and only showed up because I rebased and re-ran rather than trusting two green branches to compose.

PR #894, opened 2026-08-17
feat(testing): bootstrap one module with its neighbours doubled, and assert boundaries from a test

### LINKED ISSUE #878: Module test slices: bootstrapModule() with doubles, and boundary assertions in PHPUnit
Testing one module with its neighbours stubbed is the everyday test of a modular monolith, and Gacela has every primitive for it but no one-call way to do it, and no PHPUnit-side way to assert the module's boundaries. Spring Modulith ships `@ApplicationModuleTest` (bootstrap one module, mock or bootstrap its declared dependencies) and ArchUnit-style assertions in the test suite; Laravel has `$this->mock()` / `swap()`; Gacela makes you assemble it by hand each time.

## What exists today

- `Gacela::overrideExistingResolvedClass(BillingFacade::class, $mock)` — an **instance** double for anything the `Locator` resolves (`src/Framework/Container/Locator.php` consults `AnonymousGlobal` first). This already covers a neighbour Facade obtained through `#[ServiceMap]` or `getProvidedDependency()`.
- `ContainerFixture::swapModuleFactory/Config/Provider(facadeClass, $double)` (`src/Framework/Testing/ContainerFixture.php`) — pillar doubles, which need a whole `AbstractFactory` subclass written.
- `GacelaTestCase::bootstrapGacela()`, `assertServiceResolved()`, `assertBindingRegistered()`, `recordedGacelaEvents()`.
- `debug:graph --check --rules rules.json` and `DeclaredModuleDependencyAnalyser` — module dependency rules, but only reachable from the CLI and the static-analysis host, not from a test method.

So the capability is *possible*; the DX is: bootstrap the whole app, remember which override API applies to which kind of dependency, and check boundaries in a different tool.

### 1. `bootstrapModule()` on `GacelaTestCase`

```php
$this->bootstrapModule(
    InvoiceFacade::class,
    doubles: [
        BillingFacade::class => $this->createStub(BillingFacade::class),
        PaymentGateway::class => new FakeGateway(),
    ],
);
```

- Resolves the module from the Facade (same derivation `swapModulePillar()` uses; the same `ModuleDoubleException::notAFacade` guard).
- Narrows `appModulePaths` to that module's directory so `doctor`/discovery-driven behaviour sees one module — the "slice".
- Applies each double through the right primitive: a Facade or Locator-resolved class → `overrideExistingResolvedClass()`; a container id / interface → `addBinding()`; a pillar double → the existing `swapModule*()`.
- Fails loudly on a double for a class the module never depends on (typo, or a dependency that has since been removed): read the module's Provider `#[Provides]` and `#[ServiceMap]` declarations, which `debug:module` already inspects.

### 2. Boundary assertions in PHPUnit

```php
self::assertModuleDependsOnlyOn(InvoiceFacade::class, [BillingFacade::class, CustomerFacade::class]);
self::assertNoModuleCycles();
```

Read the same graph and the same rules file `debug:graph --check` reads (`src/Console/...Graph...`), so the CLI gate and the test agree by construction. This is what makes a boundary decision live next to the module's own tests instead of only in CI configuration.

### 3. `recordedGacelaEventsOf()` for a project's own events

Follow-up to #880: once modules dispatch their own events, the recording listener already installed by `bootstrapGacela()` should see them, so `assertEventDispatched(InvoiceIssued::class)` is one line.

## Design calls

- `bootstrapModule()` bootstraps once per test like `bootstrapGacela()`; the `tearDown` already resets. Do not introduce a second lifecycle.
- The doubles map is keyed by class-string; the value is `object|callable|class-string`. Same shape `addBinding()` accepts, so nothing new to learn.
- Arch assertions are static methods on `GacelaTestCase` and also on a standalone `ModuleAssertions` trait, so a project with its own base test class can use them.
- Nothing here needs PHPUnit internals beyond `Assert::assertThat`; keep the `phpunit/phpunit` dependency where it is (dev + suggested).

## Acceptance

- A test for `Invoice` with `Billing` stubbed passes without `Billing`'s Factory or Provider being loaded (assert via `recordedGacelaEventsOf(ResolvedClassCreatedEvent::class)`).
- `assertModuleDependsOnlyOn()` fails on the same edge `debug:graph --check` fails on, with a message naming the file and line that creates it.
- `docs/testing.md` (or the existing testing section) shows the slice pattern first, the pillar doubles second.

Chemaclass:
## 📚 Description

Testing one module with its neighbours stubbed is the everyday test of a modular monolith. Gacela had every primitive for it — `overrideExistingResolvedClass()`, `swapModuleFactory/Config/Provider()`, `setAppModulePaths()` — and no one call that puts them together, so every project assembled the slice by hand and had to remember which override API applies to which kind of dependency. Module boundaries were checkable too, but only from the CLI and the static analysers, never from a test method.

Closes #878

## 🔖 Changes

- **`GacelaTestCase::bootstrapModule(FacadeClass, doubles: [...])`** — bootstraps one module with its neighbours replaced. It derives the module from the Facade (same derivation and the same `ModuleDoubleException::notAFacade` guard the pillar doubles already use), narrows `setAppModulePaths()` to that module so discovery sees one, and routes each double through the right primitive by what it is: a pillar instance → the existing `swapModule*()`, anything the `Locator` resolves → `overrideExistingResolvedClass()`, an interface or container id → `addBinding()`. One lifecycle, the existing `tearDown` resets it
- **`Gacela\Console\Testing\ModuleAssertions`** — `assertModuleDependsOnlyOn()`, `assertNoModuleCycles()`, and the declared-rules assertion, over the same `ModuleGraphBuilder`/`ModuleRuleChecker`/`CycleAllowList` that `debug:graph --check` uses, so a boundary holds in the test and in CI or in neither
- **The import parser now carries the declaring line**, so a failure names the edge *and* the file and line that creates it rather than just the modules involved
- A setup mistake — an unreadable allowed-cycles file, a rule governing no module, a module the scan never found — raises `ModuleAssertionException` rather than a test failure, so "your fixture is wrong" never reads as "your architecture is wrong"
- Demonstrated in the reference application (`InvoicingSliceTest`): `Billing` bootstrapped with `Customer` and `Notification` doubled. That the replaced modules are never built is asserted from the recorded `ResolvedClassCreatedEvent`s, not from trusting the setup
- Docs: `docs/testing.md` gains the slice pattern and the boundary assertions

### One deviation from the issue

The issue put the boundary assertions on `GacelaTestCase`. They are a **standalone trait in the Console namespace** instead. `src/Framework/` references `Gacela\Console` in zero files, the graph engine lives in `src/Console/Domain/ModuleGraph/`, and this repo gates its own layering with `composer module-cycles`. Putting them on `GacelaTestCase` would have inverted Framework → Console — an architecture regression in the framework that sells module boundaries. A project uses them with `use ModuleAssertions;` in its own base test case, which is what the issue's "standalone trait" bullet asked for anyway. `bootstrapModule()` needs only Framework primitives, so it is on `GacelaTestCase` as proposed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,22 @@ Lists every event the framework can dispatch, which of them your project listens
 - **Paths skipped** — an `appModulePaths` entry that is not a directory is reported as `Not scanned, not a directory: …`, and by a new `module paths` doctor check so `--strict` can fail on it
 - **The filter** — `doctor` names it and how many modules it matched, including none, which used to end in `All checks passed`
 
+#### Module test slices
+
+`GacelaTestCase::bootstrapModule($appRoot, InvoiceFacade::class, doubles: [...])` bootstraps one module with its neighbours replaced — the everyday test of a modular application, which needed a bootstrap plus three override APIs and the question of which one applies to which kind of dependency.
+
+- Narrows module discovery to that module's directory, so `doctor`, `list:modules` and `debug:graph` answer about one module
+- Routes each double by what it is: a pillar instance to `swapModule*()`, a class or interface to a binding, a lazy service and a resolved-class override, a container id to a service extension — the only reach a bootstrap has into a module's own Provider
+- Refuses a double that is not an instance of the class it is registered under; a Facade left `final` cannot be doubled by anyone, so replace that module's Factory instead
+
+#### Module boundaries in PHPUnit
+
+`Gacela\Console\Testing\ModuleAssertions` is a standalone trait with `assertModuleDependsOnlyOn()`, `assertNoModuleCycles()` and `assertModuleRulesHold()`.
+
+- Reads the graph, allowed-cycles file and `module-rules.json` that `debug:graph --check` reads, so a boundary holds in the test and in CI or in neither
+- Every failure names the offending edge and the `use` statement behind it, as `file:line`
+- In the `Gacela\Console` namespace because the graph is built by scanning source files, and `Gacela\Framework` must not depend on `Gacela\Console`
+
 #### Reference application
 
 An invoicing SaaS under `tests/Feature/ReferenceApp/`, wired with every feature at once and run on every pull request. Five modules, a three-layer harness — behaviour, every shipped command, and a guard that fails when a new `GacelaConfig` method, attribute or command is not used by it — and both analysers over the application with the opt-in architecture rules turned on. See [docs/reference-app.md](docs/reference-app.md).
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,7 +12,7 @@
 - [Module health checks](module-health-checks.md) — report module operational status
 - [Profiling](profiling.md) — instrument code with the in-memory `Profiler` and read it back with `profile:report`
 - [Events](events.md) — listen to Gacela internals: dispatch model, event catalog, cookbook
-- [Testing](testing.md) — `GacelaTestCase`: bootstrap isolation, config overrides, container assertions
+- [Testing](testing.md) — `GacelaTestCase`: bootstrap isolation, module slices with doubles, container and boundary assertions
 - [Caching](caching.md) — overview of Gacela's three caching layers and when to reach for each
 - [Cacheable methods](cacheable-methods.md) — cache facade method results with `#[Cacheable]`
 - [FileCache and ScopedCache](file-cache.md) — cache arbitrary application data with atomic writes and cascading invalidation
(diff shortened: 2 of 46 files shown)


Chemaclass:
Mutation gate addressed. Covered Code MSI on the changed lines went **76% → 100%** (157 mutants killed, filtered run against `origin/main`).

Most of the 48 escaped mutants had one cause rather than 48: each finding line was written twice — once for the findings list, once for the report — so no assertion on the message could ever reach the copy in the list. Every finding now has a single definition the report is built from, which also removed an `array_map`, an `array_filter` and three report builders.

Three guards were deleted rather than tested, because no input could tell them apart: the pre-bootstrap Facade guard the pillar swap re-checks immediately after, a `class_exists` test around a resolver override that is only ever looked up by class, and two `resetCache()` calls the bootstrap and the swaps had already made.

What remains ignored in `infection.json5` is per-method and carries its reason: `ProtectedVisibility` on the trait methods (a trait is flattened into the using class, so no caller can tell `protected` from `private` — the only callers are the trait's own consumers), and the `Concat` layout around the failure messages: what each message *says* is asserted — the edge, the file and line, the rule's reason, the modules the scan found — while the newline between a finding and its evidence and the `"  - "` indent are not, because pinning those is a golden master over whitespace.

